PR #1, opened 2021-06-17
encapsulate operator to accommodate testbed etc

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
Encapsulate KIT tool into KIT folder to accommodate test-bed and test-specs


prateekgogia:
Not sure if we want to name it kit, as it will stutter. Holding onto changing the name of the directory until we finalize the name for this component.

kit/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
ellistarn, line 1: Thanks for adding this. Let's do a single LICENSE file at the top of the repo. I want to avoid nesting licenses as it may make folks wary of licenses they haven't found yet in the tree.

hakuna-matatah: My understanding is that this is not a permanent place holder for the whole KIT Eco system project to co-exist in one repository. So, it will be easier to migrate to other repositories with folders containing all of it together than picking and choosing certain files from parent folder when we move to individual repositories ?  wdyt ?

ellistarn: Agree in general, minus the LICENSE file.

ellistarn:
Let's name this folder "operator"

ellistarn (approved):

prateekgogia:
Looks like the import paths were not updated for the go files in the repo.

PR #2, opened 2021-06-18
testbed integrated with flux and other addons like metrics server and tekton

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*

Testbed with EKS cluster is created and integrated with flux. With this PR basically what you out of the box is EKS with flux, tekton pipelines, tekton dashboards, tekton triggers and metrics-server installed on it. ( Bit overloaded to commit all of this in one single PR)

### DIFF
--- a/testbed/.gitignore
+++ b/testbed/.gitignore
@@ -1,0 +1,11 @@
+# CDK asset staging directory
+*.js
+!jest.config.js
+*.d.ts
+bin/*.d.ts
+bin/*.js
+node_modules
+cdk.context.json
+jest.config.js
+.cdk.staging
+cdk.out
--- a/testbed/README.md
+++ b/testbed/README.md
@@ -1,0 +1,16 @@
+ # EKS testbed with GitOps using Flux, deployed using AWS CDK
+
+More details to come 
+
+You should explore the contents of this project. It demonstrates a CDK app with an instance of a stack (`Stack`)
+
+The `cdk.json` file tells the CDK Toolkit how to execute your app.
+
+## Useful commands
+
+ * `npm run build`   compile typescript to js
+ * `npm run watch`   watch for changes and compile
+ * `npm run test`    perform the jest unit tests
+ * `cdk deploy`      deploy this stack to your default AWS account/region
+ * `cdk diff`        compare deployed stack with current state
+ * `cdk synth`       emits the synthesized CloudFormation template
--- a/testbed/addons.ts
+++ b/testbed/addons.ts
@@ -1,0 +1,19 @@
+import cdk = require('@aws-cdk/core');
+import * as eks from '@aws-cdk/aws-eks';
+import { Flux, RepositoryProps } from './addons/flux/construct';
+
+export interface AddonsProps {
+  cluster: eks.Cluster
+  repositories: RepositoryProps[];
+}
+
+export class Addons extends cdk.Construct {
+  constructor(scope: cdk.Construct, id: string, props: AddonsProps) {
+    super(scope, id);
+
+    new Flux(this, 'Flux', {
+      cluster: props.cluster,
+      repositories: props.repositories,
+    });
+  }
+}
--- a/testbed/addons/flux/construct.ts
+++ b/testbed/addons/flux/construct.ts
@@ -1,0 +1,70 @@
+import cdk = require('@aws-cdk/core');
+import eks = require('@aws-cdk/aws-eks');
+import * as yaml from 'js-yaml';
+import * as request from 'sync-request';
+
+export interface FluxProps {
+    cluster: eks.Cluster;
+    repositories: RepositoryProps[];
+}
+
+export interface RepositoryProps {
+    url: string;
+    name: string;
+    branch?: string;
+    path?: string;
+}
+
+export class Flux extends cdk.Construct {
+    constructor(scope: cdk.Construct, id: string, props: FluxProps) {
+        super(scope, id);
+
+        const fluxManifest = props.cluster.addManifest(
+            'flux', ...yaml.loadAll(request.default(
+                "GET", "https://github.com/fluxcd/flux2/releases/download/v0.15.0/install.yaml").getBody().toString()));
+
+        props.repositories.forEach(function (value) {
+            // Bootstrap manifests
+            const gitRepoManifest = props.cluster.addManifest('GitRepoSelf', {
+                apiVersion: 'source.toolkit.fluxcd.io/v1beta1',
+                kind: 'GitRepository',
+                metadata: {
+                    name: value.name,
+                    namespace: 'default'
+                },
+                spec: {
+                    // we can adjust this later if we want to be more aggressive  
+                    interval: '5m0s',
+                    ref: {
+                        branch: value.branch ?? "main",
+                    },
+                    secretRef: {
+                        name: 'github-key'
+                    },
+                    url: value.url
+                }
+            });
+            gitRepoManifest.node.addDependency(fluxManifest);
+            const kustomizationManifest = props.cluster.addManifest('KustomizationSelf', {
+                apiVersion: 'kustomize.toolkit.fluxcd.io/v1beta1',
+                kind: 'Kustomization',
+                metadata: {
+                    name: value.name,
+                    namespace: 'default'
+                },
+                spec: {
+                    // we can adjust this later if we want to be more aggressive  
+                    interval: '5m0s',
+                    path: value.path ?? "test/workflows",
+                    prune: true,
+                    sourceRef: {
+                        kind: 'GitRepository',
+                        name: value.name
+                    },
+                    validation: 'client'
+                }
+            });
+            kustomizationManifest.node.addDependency(fluxManifest);
+        });
+    }
+}
--- a/testbed/addons/kustomization.yaml
+++ b/testbed/addons/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - registry.yaml
--- a/testbed/addons/metrics-server/kustomization.yaml
+++ b/testbed/addons/metrics-server/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.4.2/components.yaml
--- a/testbed/infrastructure.ts
+++ b/testbed/infrastructure.ts
@@ -1,0 +1,45 @@
+import * as iam from '@aws-cdk/aws-iam';
+import * as eks from '@aws-cdk/aws-eks';
+import * as ec2 from '@aws-cdk/aws-ec2';
+import * as cdk from '@aws-cdk/core'
+import { StackProps } from '@aws-cdk/core';
+
+export class Infrastructure extends cdk.Construct {
+    cluster: eks.Cluster;
+    constructor(scope: cdk.Construct, id: string, props: StackProps) {
+        super(scope, id);
+
+        const vpc = new ec2.Vpc(this, 'VPC', {});
+
+        this.cluster = new eks.Cluster(this, 'Cluster', {
+            vpc: vpc,
+            role: new iam.Role(this, 'ClusterRole', {
+                assumedBy: new iam.ServicePrincipal('eks.amazonaws.com'),
+                managedPolicies: [
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSClusterPolicy'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController')
+                ]
+            }),
+            version: eks.KubernetesVersion.V1_19,
+            defaultCapacity: 0
+        });
+
+        // for time being we start with these defaults; To-do use karpenter to manage this
+        this.cluster.addNodegroupCapacity('WorkerNodeGroup', {
+            subnets: vpc.selectSubnets({
+                subnetType: ec2.SubnetType.PRIVATE
+            }),
+            nodeRole: new iam.Role(this, 'WorkerRole', {
+                assumedBy: new iam.ServicePrincipal('ec2.amazonaws.com'),
+                managedPolicies: [
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSWorkerNodePolicy'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEC2ContainerRegistryReadOnly'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKS_CNI_Policy'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController') // Allows us to use Security Groups for pods
+                ]
+            }),
+            minSize: 5,
+            maxSize: 20
+        });
+    }
+}
--- a/testbed/stack.ts
+++ b/testbed/stack.ts
@@ -1,0 +1,9 @@
+import * as cdk from '@aws-cdk/core'
+import { Testbed, TestbedProps } from './construct';
+
+export class TestbedStack extends cdk.Stack {
+  constructor(scope: cdk.Construct, id: string, props: TestbedProps) {
+    super(scope, id, props);
+    new Testbed(this, "Testbed", props)
+  }
+}
--- a/testbed/addons/registry.yaml
+++ b/testbed/addons/registry.yaml
@@ -1,0 +1,59 @@
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: metrics-server
+  namespace: default
+spec:
+  interval: 5m0s
+  path: ./kubernetes-iteration-toolkit/testbed/addons/metrics-server
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: default
+  validation: client
+
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: tekton-pipelines
+  namespace: default
+spec:
+  interval: 5m0s
+  path: ./kubernetes-iteration-toolkit/testbed/addons/tekton-pipelines
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: default
+  validation: client
+
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: tekton-triggers
+  namespace: default
+spec:
+  interval: 5m0s
+  path: ./kubernetes-iteration-toolkit/testbed/addons/tekton-triggers
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: default
+  validation: client
+
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: tekton-dashboard
+  namespace: default
+spec:
+  interval: 5m0s
+  path: ./kubernetes-iteration-toolkit/testbed/addons/tekton-dashboard
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: default
+  validation: client
--- a/testbed/addons/tekton-dashboard/kustomization.yaml
+++ b/testbed/addons/tekton-dashboard/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/tektoncd/dashboard/releases/latest/download/tekton-dashboard-release.yaml
--- a/testbed/addons/tekton-pipelines/kustomization.yaml
+++ b/testbed/addons/tekton-pipelines/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
--- a/testbed/addons/tekton-triggers/kustomization.yaml
+++ b/testbed/addons/tekton-triggers/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://storage.googleapis.com/tekton-releases/triggers/latest/release.yaml
--- a/testbed/cdk.json
+++ b/testbed/cdk.json
@@ -1,0 +1,17 @@
+{
+  "app": "npx ts-node --prefer-ts-exts bin/testbed.ts",
+  "context": {
+    "@aws-cdk/aws-apigateway:usagePlanKeyOrderInsensitiveId": true,
+    "@aws-cdk/core:enableStackNameDuplicates": "true",
+    "aws-cdk:enableDiffNoFail": "true",
+    "@aws-cdk/core:stackRelativeExports": "true",
+    "@aws-cdk/aws-ecr-assets:dockerIgnoreSupport": true,
+    "@aws-cdk/aws-secretsmanager:parseOwnedSecretName": true,
+    "@aws-cdk/aws-kms:defaultKeyPolicies": true,
+    "@aws-cdk/aws-s3:grantWriteWithoutAcl": true,
+    "@aws-cdk/aws-ecs-patterns:removeDefaultDesiredCount": true,
+    "@aws-cdk/aws-rds:lowercaseDbIdentifier": true,
+    "@aws-cdk/aws-efs:defaultEncryptionAtRest": true,
+    "@aws-cdk/aws-lambda:recognizeVersionProps": true
+  }
+}
--- a/testbed/construct.ts
+++ b/testbed/construct.ts
@@ -1,0 +1,17 @@
+import { Cluster } from '@aws-cdk/aws-eks';
+import * as cdk from '@aws-cdk/core'
+import { StackProps } from '@aws-cdk/core';
+import { Addons } from './addons';
+import { RepositoryProps } from './addons/flux/construct';
+import { Infrastructure } from './infrastructure';
+
+export interface TestbedProps extends cdk.StackProps {
+    repositories: RepositoryProps[];
+}
+
+export class Testbed extends cdk.Construct {
+    constructor(scope: cdk.Construct, id: string, props: TestbedProps) {
+        super(scope, id);
+        new Addons(this, 'addons', { cluster: new Infrastructure(this, 'eks', props).cluster, repositories: props.repositories })
+    }
+}
--- a/testbed/package-lock.json
+++ b/testbed/package-lock.json
@@ -1,0 +1,7225 @@
+{
+  "name": "testbed",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@aws-cdk/assert": {
+      "version": "1.108.1",
+      "resolved": "https://registry.npmjs.org/@aws-cdk/assert/-/assert-1.108.1.tgz",
+      "integrity": "sha512-dAeoc04S5HpHzYeyVCg15AWW+jwe+LBD7OkRxz7Nsrg+ZpsFbXWhNON9Fvtzsef7G1ztKIBAFreD/jy7zrPSMA==",
+      "dev": true,
+      "requires": {
+        "@aws-cdk/cloudformation-diff": "1.108.1",
+        "@aws-cdk/core": "1.108.1",
+        "@aws-cdk/cx-api": "1.108.1",
+        "constructs": "^3.3.69"
+      }
+    },
+    "@aws-cdk/assets": {
+      "version": "1.108.1",
+      "resolved": "https://registry.npmjs.org/@aws-cdk/assets/-/assets-1.108.1.tgz",
+      "integrity": "sha512-SNj0y944vstV98qfHKe+lxuLgRLp2hMeCa8aKZPH95MfWqJ5ioNp/cP3GIF0j8Ek0NsREgsN7SzsjYTR2vccdw==",
+      "requires": {
+        "@aws-cdk/core": "1.108.1",
+        "@aws-cdk/cx-api": "1.108.1",
+        "constructs": "^3.3.69"
+      }
+    },
+    "@aws-cdk/aws-applicationautoscaling": {
+      "version": "1.108.1",
+      "resolved": "https://registry.npmjs.org/@aws-cdk/aws-applicationautoscaling/-/aws-applicationautoscaling-1.108.1.tgz",
+      "integrity": "sha512-75iGnGBdb3cXE/SnpTDNh+LCOnAHaK0GEwMGgxHOr0IXbMYbLRuRHQdSjHffRzTF9zH17DoW+fi9qvSkkFduKw==",
+      "requires": {
+        "@aws-cdk/aws-autoscaling-common": "1.108.1",
+        "@aws-cdk/aws-cloudwatch": "1.108.1",
+        "@aws-cdk/aws-iam": "1.108.1",
+        "@aws-cdk/core": "1.108.1",
+        "constructs": "^3.3.69"
+      }
+    },
+    "@aws-cdk/aws-autoscaling": {
+      "version": "1.108.1",
+      "resolved": "https://registry.npmjs.org/@aws-cdk/aws-autoscaling/-/aws-autoscaling-1.108.1.tgz",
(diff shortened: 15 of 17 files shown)


testbed/.gitignore
```diff
@@ -0,0 +1,13 @@
+*.js
+!jest.config.js
+*.d.ts
+node_modules
+.DS_Store
```
ellistarn, line 5: I prefer to keep .gitignores minimal and intentional and specific to the files necessary for the project (e.g. node_modules should be in here). 

In general, it's best practice to have a developer global ~/.gitignore file that ignores things like IDE settings and OSX finder registries (.vscode, .DS_STORE). Otherwise, the .gitignore must contain the union set of every developers environment, which can get out of hand.

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
+import cdk = require('@aws-cdk/core');
```
ellistarn, line 1: let's avoid including versioning concepts as part of files names (e.g. fluxv2.ts -> flux.ts)

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as request from 'sync-request';
+
+export interface FluxV2Props extends cdk.StackProps {
+    cluster: eks.Cluster;
+    fluxRepos: Repository[];
+    secretName: string;
+    fluxVersion: string;
+}
+export class Repository {
+    repoUrl: string;
```
ellistarn, line 13: These variable names are already contained within a concept we called "repository". Naming variables with "repo", causes us to "stutter". For example, `repository.repo___`. Consider `repository.branch`, `repository.url`.

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+    repoBranch: string;
+    repoPath: string;
+
+    constructor(repoUrl: string, repoBranch?: string, repoPath?: string) {
+        if (!repoBranch) {
```
ellistarn, line 18: Check out typescript's defaulting syntax
`this.branch = this.branch ?? "main"`

https://stackoverflow.com/questions/47328862/how-to-set-a-variable-if-undefined-in-typescript

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+    fluxRepos: Repository[];
+    secretName: string;
+    fluxVersion: string;
+}
+export class Repository {
+    repoUrl: string;
+    repoBranch: string;
+    repoPath: string;
+
+    constructor(repoUrl: string, repoBranch?: string, repoPath?: string) {
```
ellistarn, line 17: Instead of passing these variables as strings, which can introduce readability challenges from the caller side, use an options class pattern.

e.g. 
```
new Repository("foo", "bar", "baz") # <-- What are these variables?
```
vs 
```
new Repository({
  url: "...",
  branch: "...",
})
```

export class RepositoryOptions{
  branch: string
  path: string
  url: string
}

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+                        kind: 'GitRepository',
+                        name: 'flux-system'+index
+                    },
+                    validation: 'client'
+                }
+            });
+            kustomizationManifest.node.addDependency(fluxManifest);
+        }
+    }
+}
```
ellistarn, line 110: nit newline

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export class FluxV2 extends cdk.Construct {
+    constructor(scope: cdk.Construct, id: string, props: FluxV2Props) {
+        super(scope, id);
+
+        // Actually install Flux components onto the cluster
+        const fluxRelease = new FluxRelease(props.fluxVersion);
+        const fluxManifest = props.cluster.addManifest('fluxManifest', ...fluxRelease.getManifest());
+
+        for (let index = 0; index < props.fluxRepos.length; index++) {
```
ellistarn, line 63: A couple ways to clean this up. Add a function to this class like `addManifest()`. I like the parity here w/ cluster.addManifest as well.

Then in your constructor code `props.repositories.foreach(this.addManifest)`. This will keep your constructor well encapsulated. Typically, if you're finding yourself writing big for loops, it's a good hint that its time to break the logic apart.

testbed/eks/lib/addons/fluxv2.ts
ellistarn: Also -- consider creating one package per addon, since we might end up having many files in support of the addon. Off the top of my head, we'll want to separate cloud formation templates and typescript code. If all addons are in the same package, this is going to get unwieldy.

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+        this.manifestUrl = `https://github.com/fluxcd/flux2/releases/download/${this.fluxVersion}/install.yaml`
+    }
+
+    public getUrl(): string {
+        return this.manifestUrl;
+    }
+
+    public getManifest(): any {
+        this.installManifest = yaml.loadAll(
+            request.default('GET', this.manifestUrl)
```
ellistarn, line 47: consider using helm to install flux `https://fluxcd.io/legacy/flux/tutorials/get-started-helm/#install-flux`

ellistarn, line 47: The eks construct docs recommend `const manifest = yaml.safeLoadAll(request('GET', manifestUrl).getBody());` which I think is pretty clean.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as ec2 from '@aws-cdk/aws-ec2';
+import * as cdk from '@aws-cdk/core'
+import { FluxV2, Repository } from './addons/fluxv2';
+
+export interface TestbedProps extends cdk.StackProps {
+  fluxRepos: Repository[];
+  fluxVersion: string;
+}
+
+export class TestbedStack extends cdk.Stack {
```
ellistarn, line 12: Thoughts on just moving this whole thing up to `/testbed/stack.ts` to avoid layers of indirection and stuttering? I'd personally toss any notion of `lib`, since I don't expect us to be writing many tests. 

I'd also separate this stack and its properties like fluxVersion and repositories into a single file, and then build a top level construct that encapsulates the rest of the things. Not only is this a cleaner separation of concerns, it also allows allows customers to define their own stack that consumes this construct, since you can't wrap a stack within another stack.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+  constructor(scope: cdk.Construct, id: string, props: TestbedProps) {
+    super(scope, id, props);
+
+    // A VPC, including NAT GWs, IGWs, where we will run our cluster
+    const vpc = new ec2.Vpc(this, 'VPC', {});
+
+    // The IAM role that will be used by EKS
+    const clusterRole = new iam.Role(this, 'ClusterRole', {
```
ellistarn, line 20: Consider inlining

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+    // The EKS cluster
+    const cluster = new eks.Cluster(this, 'Cluster', {
+      vpc: vpc,
+      role: clusterRole,
+      version: eks.KubernetesVersion.V1_19,
+      defaultCapacity: 0
+    });
+
+    // Worker node IAM role
+    const workerRole = new iam.Role(this, 'WorkerRole', {
```
ellistarn, line 37: consider inlining

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Select the private subnets created in our VPC and place our worker nodes there
+    const privateSubnets = vpc.selectSubnets({
+      subnetType: ec2.SubnetType.PRIVATE
+    });
+
+    // for time being we start with these defaults; To-do use karpenter to manage this
+    cluster.addNodegroupCapacity('WorkerNodeGroup', {
+      subnets: privateSubnets,
+      nodeRole: workerRole,
+      minSize: 3,
```
ellistarn, line 56: Why [3, 20]?

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+      managedPolicies: [
+        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSWorkerNodePolicy'),
+        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEC2ContainerRegistryReadOnly'),
+        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKS_CNI_Policy'),
+        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController') // Allows us to use Security Groups for pods
+      ]
+    });
+
+    // Select the private subnets created in our VPC and place our worker nodes there
+    const privateSubnets = vpc.selectSubnets({
```
ellistarn, line 48: consider inlining

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // for time being we start with these defaults; To-do use karpenter to manage this
+    cluster.addNodegroupCapacity('WorkerNodeGroup', {
+      subnets: privateSubnets,
+      nodeRole: workerRole,
+      minSize: 3,
+      maxSize: 20
+    });
+
+    // Add FluxV2
```
ellistarn, line 60: I'd love to encapsulate this into an Addons construct so that we can hide these details.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+    new FluxV2(this, 'FluxV2', {
+      cluster: cluster,
+      secretName: 'github-key',
+      fluxRepos: props.fluxRepos,
+      fluxVersion: props.fluxVersion
+    });
+
+  }
+}
```
ellistarn, line 69: nit: newline. Your editor should be able to automate this for ya.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+      minSize: 3,
+      maxSize: 20
+    });
+
+    // Add FluxV2
+    new FluxV2(this, 'FluxV2', {
+      cluster: cluster,
+      secretName: 'github-key',
+      fluxRepos: props.fluxRepos,
+      fluxVersion: props.fluxVersion
```
ellistarn, line 65: Consider a lazy parameterization strategy rather than a preemptive one. I'm not seeing the use case where folks will need to control the version of flux. I'd rather we map KIT versions to a set of dependency versions, as if it were a typical build time dependency.

testbed/eks/package.json
```diff
@@ -0,0 +1,33 @@
+{
```
ellistarn, line 1: I'm not seeing the value of this living inside and eks folder. Thoughts on just moving this to the top level of the testbed?

testbed/eks/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "testbed",
+  "version": "0.1.0",
+  "bin": {
+    "testbed": "bin/testbed.js"
+  },
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc -w",
+    "test": "jest",
```
ellistarn, line 10: We can probably drop this.

testbed/eks/test/testbed.test.ts
```diff
@@ -0,0 +1,13 @@
+import { expect as expectCDK, matchTemplate, MatchStyle } from '@aws-cdk/assert';
```
ellistarn, line 1: I'd hope we don't have any logic in our stack, which means we shouldn't need to test anything.

testbed/k8s-config/addons.yaml
```diff
@@ -0,0 +1,62 @@
...
+    kind: GitRepository
+    name: flux-system0
+  validation: client
+
+---
+
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
+metadata:
+  name: tekton-triggers
```
ellistarn, line 37: Do we need triggers?

testbed/k8s-config/addons.yaml
```diff
@@ -0,0 +1,62 @@
+---
```
ellistarn, line 1: Consider encapsulating all of this into addons e.g.
```
testbed/addons/metrics-server
testbed/addons/tekton

```

testbed/eks/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "testbed",
+  "version": "0.1.0",
+  "bin": {
+    "testbed": "bin/testbed.js"
```
ellistarn, line 5: I think this file may no longer exist :)

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    // Select the private subnets created in our VPC and place our worker nodes there
+    const privateSubnets = vpc.selectSubnets({
+      subnetType: ec2.SubnetType.PRIVATE
+    });
+
+    // for time being we start with these defaults; To-do use karpenter to manage this
+    cluster.addNodegroupCapacity('WorkerNodeGroup', {
+      subnets: privateSubnets,
```
ellistarn, line 54: +1 I like that we're using private subnets.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+        iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController')
+      ]
+    });
+
+    // The EKS cluster
+    const cluster = new eks.Cluster(this, 'Cluster', {
+      vpc: vpc,
+      role: clusterRole,
+      version: eks.KubernetesVersion.V1_19,
+      defaultCapacity: 0
```
ellistarn, line 33: +1 I like that we make this explicit.

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
+import * as iam from '@aws-cdk/aws-iam';
+import * as eks from '@aws-cdk/aws-eks';
+import * as ec2 from '@aws-cdk/aws-ec2';
+import * as cdk from '@aws-cdk/core'
+import { FluxV2, Repository } from './addons/fluxv2';
+
+export interface TestbedProps extends cdk.StackProps {
+  fluxRepos: Repository[];
```
ellistarn, line 8: flux is an implementation detail, thoughts on just `repositories`

testbed/eks/lib/testbed-stack.ts
```diff
@@ -0,0 +1,69 @@
+import * as iam from '@aws-cdk/aws-iam';
+import * as eks from '@aws-cdk/aws-eks';
+import * as ec2 from '@aws-cdk/aws-ec2';
+import * as cdk from '@aws-cdk/core'
+import { FluxV2, Repository } from './addons/fluxv2';
+
+export interface TestbedProps extends cdk.StackProps {
```
ellistarn, line 7: The cdk ecosystem uses both terms "props" and "options": https://github.com/cdk8s-team/cdk8s/blob/master/examples/typescript/podinfo/lib/ingress.ts#L36. I'm a bit partial to props, but happy to go either way.

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+        if (!repoPath) {
+            this.repoPath = "test/workflows"
+        } else {
+            this.repoPath = repoPath
+        }
+        this.repoUrl = repoUrl
+    }
+}
+class FluxRelease {
```
ellistarn, line 31: I'm not sure this FluxRelease class is worth its weight. Then entire class can be boiled down into 
```
props.cluster.addManifest('flux', ...yaml.safeLoadAll(
  request('GET', `https://github.com/fluxcd/flux2/releases/download/${this.fluxVersion}/install.yaml).getBody()
))
```
Consider putting this directly into the Flux construct. This might be further simplified with using helm (though it looks like you might still need something like this to install the CRDs).

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+                    name: 'flux-system' + index,
+                    namespace: 'flux-system'
+                },
+                spec: {
+                    // we can adjust this later if we want to be more aggressive  
+                    interval: '5m0s',
+                    ref: {
+                        branch: props.fluxRepos[index].repoBranch,
+                    },
+                    secretRef: {
```
ellistarn, line 80: consider inlining a lot of this short stuff for brevity
```
secretRef: { name: props.secretName }
```

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+                        name: props.secretName
```
ellistarn, line 81: consider just hardcoding this, rather than building configurability for it.

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+                    },
+                    url: props.fluxRepos[index].repoUrl
+                }
+            });
+            gitRepoManifest.node.addDependency(fluxManifest);
```
ellistarn, line 86: consider using the KubernetesManifest struct for declarative style instead of imperative style. This enables us to leverage variadics and establish the dependency more cleanly.

```
const cluster = new EKSCluster(...)
const flux = new Flux(...)

...

const manifest = new KubernetesManifest(..., {
  cluster: cluster,
  manifest: [ ... ] # Both Repo and Kustomization can be defined in this variadic
}
manifest.node.addDependency(flux)
```

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+            const kustomizationManifest = props.cluster.addManifest('KustomizationSelf', {
+                apiVersion: 'kustomize.toolkit.fluxcd.io/v1beta1',
+                kind: 'Kustomization',
+                metadata: {
+                    //to-do use some meaningfulname
+                    name: 'flux-system' + index,
+                    namespace: 'flux-system'
```
ellistarn, line 93: I don't think you want to install these repositories into the flux-system namespace. Typically, foo-system namespaces are reserved for the operator, and CRD instances are defined in another namespace. In our case, I think the default namespace would be appropriate (for now).

testbed/eks/lib/addons/fluxv2.ts
```diff
@@ -0,0 +1,110 @@
...
+                    url: props.fluxRepos[index].repoUrl
+                }
+            });
+            gitRepoManifest.node.addDependency(fluxManifest);
+            const kustomizationManifest = props.cluster.addManifest('KustomizationSelf', {
+                apiVersion: 'kustomize.toolkit.fluxcd.io/v1beta1',
+                kind: 'Kustomization',
+                metadata: {
+                    //to-do use some meaningfulname
+                    name: 'flux-system' + index,
```
ellistarn, line 92: consider naming this based off of the repository name itself. You may need to sanitize it to make it k8s safe (e.g. github.com/awslabs/testbed -> github.com-awslabs-testbed)

ellistarn (changes requested):
Nice work. Let me know if you want to discuss any of the comments in person. Overall, a great starting point for our testbed work. Looking forward to seeing it in action.

testbed/README.md
```diff
@@ -0,0 +1,3 @@
+# EKS with GitOps using Flux, deployed using AWS CDK
```
prateekgogia, line 1: `EKS` -> `Kubernetes testbed`?

testbed/eks/README.md
```diff
@@ -0,0 +1,14 @@
...
+The `cdk.json` file tells the CDK Toolkit how to execute your app.
+
+## Useful commands
+
+ * `npm run build`   compile typescript to js
+ * `npm run watch`   watch for changes and compile
+ * `npm run test`    perform the jest unit tests
+ * `cdk deploy`      deploy this stack to your default AWS account/region
+ * `cdk diff`        compare deployed stack with current state
+ * `cdk synth`       emits the synthesized CloudFormation template
```
prateekgogia, line 14: Do we need to run these commands to deploy the testbed, or are these more of good to know commands?

testbed/eks/lib/addons/fluxv2.ts
prateekgogia: huge +1 on this. We should be modeling all of these as layers and these layers should be easily swapped as the add-ons might change, config might change, etc.

testbed/eks/lib/addons/fluxv2.ts
prateekgogia: oh man! I need to learn typescript!

testbed/eks/lib/addons/fluxv2.ts
prateekgogia: Also, do we need the `lib` directory under eks? Can it be just `eks/addons`?

testbed/eks/lib/addons/fluxv2.ts
hakuna-matatah: haha! I'm new to typescript as well. This is my first time writing typescript lol

testbed/eks/README.md
hakuna-matatah: Yes, we need them.

testbed/eks/lib/addons/fluxv2.ts
hakuna-matatah: Yup, makes sense. Similar to what I did for top level class Flux here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/testbed/testbed/eks/lib/testbed-stack.ts#L61-L65

https://github.com/awslabs/kubernetes-iteration-toolkit/blob/testbed/testbed/eks/lib/addons/fluxv2.ts#L6-L11

testbed/eks/lib/addons/fluxv2.ts
hakuna-matatah: I don't think encapsulating in `addManifest` function adds any value here as it's only served once.

testbed/addons.ts
```diff
@@ -0,0 +1,19 @@
+import cdk = require('@aws-cdk/core');
+import * as eks from '@aws-cdk/aws-eks';
+import { Flux, Repository } from './addons/flux/construct';
+
+export interface AddonsProps extends cdk.StackProps {
```
ellistarn, line 5: I don't think this should extend stack props.

testbed/addons.ts
```diff
@@ -0,0 +1,19 @@
...
+export class Addons extends cdk.Construct {
+  constructor(scope: cdk.Construct, id: string, props: AddonsProps) {
+    super(scope, id);
+    // Add FluxV2
+    new Flux(this, 'Flux', {
+      cluster: props.cluster,
+      repositories: props.repositories,
+    });
+  }
+}
```
ellistarn, line 19: Can you make sure your editor is inserting newlines at the bottom of files automatically? This helps with readability.

testbed/addons.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface AddonsProps extends cdk.StackProps {
+  cluster: eks.Cluster
+  repositories: Repository[];
+}
+
+export class Addons extends cdk.Construct {
+  constructor(scope: cdk.Construct, id: string, props: AddonsProps) {
+    super(scope, id);
+    // Add FluxV2
```
ellistarn, line 13: Unnecessary comment

testbed/addons.ts
```diff
@@ -0,0 +1,19 @@
...
+    new Flux(this, 'Flux', {
```
ellistarn, line 14: I think you should be able to pass props directly, since we're using the same fields 
`new Flux(this, 'Flux', props)`

testbed/addons/flux/construct.ts
```diff
@@ -0,0 +1,85 @@
+import cdk = require('@aws-cdk/core');
+import eks = require('@aws-cdk/aws-eks');
+import * as yaml from 'js-yaml';
+import * as request from 'sync-request';
+import { KubernetesManifest } from '@aws-cdk/aws-eks';
+
+export interface FluxProps extends cdk.StackProps {
```
ellistarn, line 7: Likewise

testbed/addons.ts
hakuna-matatah: why so ?

testbed/addons/flux/construct.ts
```diff
@@ -0,0 +1,85 @@
...
+    cluster: eks.Cluster;
+    repositories: Repository[];
+}
+
+export interface RepositoryOptions {
+    branch?: string;
```
ellistarn, line 13: consider putting the optional fields at the bottom.

testbed/addons/flux/construct.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as yaml from 'js-yaml';
+import * as request from 'sync-request';
+import { KubernetesManifest } from '@aws-cdk/aws-eks';
+
+export interface FluxProps extends cdk.StackProps {
+    cluster: eks.Cluster;
+    repositories: Repository[];
+}
+
+export interface RepositoryOptions {
```
ellistarn, line 12: Let's pick a consistent term `Options` or `Props`

testbed/addons/flux/construct.ts
```diff
@@ -0,0 +1,85 @@
...
+    branch?: string;
+    path?: string;
+    url: string;
+    name: string;
+}
+
+export class Repository {
```
ellistarn, line 19: I wonder if would simplify things to just remove this class entirely and instead use solely the interface. The consumer (flux construct) can do the ?? defaulting inline.

testbed/addons.ts
hakuna-matatah: my default formatter in vscode for ts apparently doesn't add it lol, i think in fact it removes it :|

testbed/addons/flux/construct.ts
```diff
@@ -0,0 +1,85 @@
...
+        this.url = options.url
+        this.name = options.name
+    }
+}
+
+export class Flux extends cdk.Construct {
+    constructor(scope: cdk.Construct, id: string, props: FluxProps) {
+        super(scope, id);
+
+        // Actually install Flux components onto the cluster
```
ellistarn, line 36: I'd avoid words like "actually".

In general I like to name things such that documentation isn't necessary (i.e. self documenting code). However, sometimes this isn't possible given the names of other components. In these cases, consider a documentation pattern of 

```
// 1. Do the first step
... 
// 2. Do another step
....
// 3. Do this one as well
...
```

testbed/addons/frill.yaml
```diff
@@ -0,0 +1,59 @@
+---
```
ellistarn, line 1: Why is this file called frill?

testbed/addons/frill.yaml
```diff
@@ -0,0 +1,59 @@
...
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta1
+kind: Kustomization
```
ellistarn, line 3: IIUC, you don't actually need to use Kustomize for a Kustomization CRD (i.e., you can skip the kustomize.yaml and just do raw YAML).

testbed/infrastructure.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as ec2 from '@aws-cdk/aws-ec2';
+import * as cdk from '@aws-cdk/core'
+import { StackProps } from '@aws-cdk/core';
+
+export class Infrastructure extends cdk.Construct {
+    cluster: eks.Cluster;
+    constructor(scope: cdk.Construct, id: string, props: StackProps) {
+        super(scope, id);
+
+        // A VPC, including NAT GWs, IGWs, where we will run our cluster
```
ellistarn, line 12: I'd argue that this is self explanatory.

testbed/infrastructure.ts
```diff
@@ -0,0 +1,47 @@
...
+        const vpc = new ec2.Vpc(this, 'VPC', {});
+
+        // The EKS cluster
```
ellistarn, line 15: likewise.

testbed/infrastructure.ts
```diff
@@ -0,0 +1,47 @@
...
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSClusterPolicy'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController')
+                ]
+            }),
+            version: eks.KubernetesVersion.V1_19,
+            defaultCapacity: 0
+        });
+
+        // for time being we start with these defaults; To-do use karpenter to manage this
+        this.cluster.addNodegroupCapacity('WorkerNodeGroup', {
```
ellistarn, line 30: Any reason to not chain directly onto line 27?

testbed/stack.ts
```diff
@@ -0,0 +1,9 @@
+import * as cdk from '@aws-cdk/core'
+import { Testbed, TestbedProps } from './construct';
+
+export class TestbedStack extends cdk.Stack {
+  constructor(scope: cdk.Construct, id: string, props: TestbedProps) {
+    super(scope, id, props);
+    new Testbed(this, "testbed", props)
```
ellistarn, line 7: I see a convention of "Testbed" for components elsewhere. Why lowercase "testbed"?

testbed/addons/metrics-server/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.4.2/components.yaml
```
ellistarn, line 4: extremely cool how this wires up arbitrary yaml using kustomize.

testbed/addons/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - frill.yaml
```
ellistarn, line 4: consider addons.yaml

testbed/addons.ts
hakuna-matatah: I think I want to leave this separate, as we add more addons we might end up having more data to pass arround in Addons interface that is not just specific to Flux ? Thoughts ?

testbed/addons/flux/construct.ts
hakuna-matatah: props it is.

testbed/addons/frill.yaml
hakuna-matatah: I'm open to suggestions :D

testbed/infrastructure.ts
hakuna-matatah: If you are asking if I can chain our customized workernode group stuff into cluster creation above, it's not possible AFSIU, only possibility is to choose defaults which I don't think I want at this point.

testbed/addons.ts
ellistarn: Any chance you can fix this setting?

ellistarn: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

testbed/addons.ts
ellistarn: Typescript will automatically strip the extra properties.

testbed/infrastructure.ts
```diff
@@ -0,0 +1,45 @@
...
+        this.cluster = new eks.Cluster(this, 'Cluster', {
+            vpc: vpc,
+            role: new iam.Role(this, 'ClusterRole', {
+                assumedBy: new iam.ServicePrincipal('eks.amazonaws.com'),
+                managedPolicies: [
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSClusterPolicy'),
+                    iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSVPCResourceController')
+                ]
+            }),
+            version: eks.KubernetesVersion.V1_19,
```
ellistarn, line 23: Thoughts on just letting this default?

ellistarn (approved):
One last comment. Nice work!

testbed/infrastructure.ts
hakuna-matatah: that hides the readability when someone wants to know in history for a given commit what the k8s version is ? We can trace it through library version though but not in our codebase. wdyt ?

ellistarn: The problem is that this code will stop working in 6 months when EKS doesn't support v1.19 anymore.

hakuna-matatah: looks like version is mandatory per cdk lib

PR #3, opened 2021-06-23
Fix import paths and remove old code

prateekgogia:
*Description of changes:*
This PR fixes the import path from old repo and removed some old code for build to work. Next PR will add Makefile and build pipeline in Github.


ellistarn (approved):
Why are the paths kit and not kubernetes-iteration-toolkit?

prateekgogia:
> Why are the paths kit and not kubernetes-iteration-toolkit?

Had a discussion offline, `kubernetes-iteration-toolkit` makes the import path too long

PR #8, opened 2021-07-07
karpenter addon for testbed

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
Integrate karpenter addon for testbed.


testbed/addons.ts
```diff
@@ -2,6 +2,7 @@ import cdk = require('@aws-cdk/core');
 import * as eks from '@aws-cdk/aws-eks';
 import { Flux, RepositoryProps } from './addons/flux/construct';
 import { AWSLoadBalancerController } from './addons/awslb/construct';
+import { KarpenterController } from './addons/karpenter/construct';
```
ellistarn, line 5: Consider just naming this Karpenter

testbed/addons/karpenter/construct.ts
```diff
@@ -0,0 +1,61 @@
...
+    namespace: string;
+}
+
+export class KarpenterController extends cdk.Construct {
+    constructor(scope: cdk.Construct, id: string, props: KarpenterProps) {
+        super(scope, id);
+
+        const sa = props.cluster.addServiceAccount('karpenter-controller-sa', {
+            name: "karpenter",
+            namespace: props.namespace
```
ellistarn, line 16: Consider defaulting this to karpenter using ?? "karpenter" so the caller doesn't always need to specify this.

testbed/addons/karpenter/construct.ts
```diff
@@ -0,0 +1,61 @@
...
+            ]
+        }),
+            {
+                username: `system:node:{{EC2PrivateDNSName}}`,
+                groups: ['system:bootstrappers', 'system:nodes']
+            }
+        )
+
+        const chart = props.cluster.addHelmChart('KarpenterHelmChart', {
+            chart: 'karpenter-v0.2.8.tgz',
```
ellistarn, line 44: Are you sure you need to put .tgz in here? Can we just call this "karpenter" and the tgz will be looked up by the version?

testbed/addons/karpenter/construct.ts
```diff
@@ -0,0 +1,61 @@
+import cdk = require('@aws-cdk/core');
+import eks = require('@aws-cdk/aws-eks');
+import iam = require('@aws-cdk/aws-iam');
+
+export interface KarpenterProps {
+    cluster: eks.Cluster;
+    namespace: string;
+}
+
+export class KarpenterController extends cdk.Construct {
```
ellistarn, line 10: This is sweet! I wonder how much effort this would be to toss this into `github.com/awslabs/karpenter/cdk`. Maybe a bad idea, since it's aws specific. We should do this once we separate the AWS cloud provider.

testbed/addons/karpenter/construct.ts
```diff
@@ -0,0 +1,61 @@
...
+        const sa = props.cluster.addServiceAccount('karpenter-controller-sa', {
+            name: "karpenter",
+            namespace: props.namespace ?? "karpenter"
+        });
+
+        sa.role.attachInlinePolicy(new iam.Policy(this, 'karpenter-controller-policy', {
+            statements: [
+                new iam.PolicyStatement({
+                    resources: ['*'],
+                    actions: ["ec2:CreateLaunchTemplate", "ec2:CreateFleet", "ec2:RunInstances", "ec2:CreateTags", "iam:PassRole", "ec2:TerminateInstances", "ec2:DescribeLaunchTemplates", "ec2:DescribeInstances", "ec2:DescribeSecurityGroups", "ec2:DescribeSubnets", "ec2:DescribeInstanceTypes", "ec2:DescribeInstanceTypeOfferings", "ec2:DescribeAvailabilityZones", "ssm:GetParameter"],
```
ellistarn, line 23: consider spreading this over multiple lines, as this line is very long.

testbed/addons/karpenter/construct.ts
```diff
@@ -0,0 +1,61 @@
...
+            }
+        )
+
+        props.cluster.addHelmChart('KarpenterHelmChart', {
+            chart: 'karpenter',
+            release: 'karpenter',
+            version: 'v0.2.8',
+            repository: 'https://awslabs.github.io/karpenter/charts',
+            namespace: props.namespace,
+            createNamespace: true,
```
ellistarn, line 49: Doesn't this fail, since the namespace is already created when you create the service account in the karpenter namespace above?

hakuna-matatah: I think this won't fail. AFASIK, it's an idempotent operation. However, namespace doesn't get automatically created when creating sa.

ellistarn (approved):

PR #9, opened 2021-07-08
Update the controlplane CRD spec

prateekgogia:
*Issue #, if available:*

*Description of changes:*
- Updates the controlplane CRD spec to include `v1.Pod` to configure control plane components like scheduler, KCM, etcd etc.
- Rename `cmd/controller` to be `cmd/operator`


operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,39 @@ type ControlPlaneList struct {
...
 }
 
-// Config provides a generic way to pass in args and images to master and etcd
+// Deployment provides a generic way to pass in args and images to master and etcd
 // components. If a user wants to change the QPS they need to provide the
 // following flag with the desired value -`kube-api-qps:100` in the args.
-type Config struct {
-	Args  map[string]string `json:"args,omitempty"`
-	Image string            `json:"image,omitempty"`
+type Deployment struct {
```
hakuna-matatah, line 69: Can't we use `type Deployment` directly from k8s lib ? what's additional significance we get from this ?

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,39 @@ type ControlPlaneList struct {
...
-
 // MasterSpec provides a way for the user to configure master instances and
 // custom flags for components running on master nodes like apiserver, KCM and
 // scheduler.
 type MasterSpec struct {
 	Instances         `json:",inline"`
-	Scheduler         *Config `json:"scheduler,omitempty"`
-	ControllerManager *Config `json:"controllerManager,omitempty"`
-	APIServer         *Config `json:"apiServer,omitempty"`
+	Scheduler         *Deployment `json:"scheduler,omitempty"`
```
hakuna-matatah, line 55: We also need other components like cert-controller and authenticator.

.github/workflows/build.yaml
```diff
@@ -23,6 +23,6 @@ jobs:
           go-version: 1.16
       - name: Build & Test
         run: |
-          go build -v ./cmd/controller
+          go build -v ./cmd/operator
```
ellistarn, line 26: I'd keep this as controller. It's a bit pedantic, but I consider the entire software installation (a set of controllers and webhooks) to together be an operator. The binary itself is the controller, which may be the core of the operator, but isn't itself the operator.

operator/config/kit-deployment.yaml
```diff
@@ -124,7 +124,7 @@ spec:
       serviceAccountName: kit
       containers:
       - name: manager
-        image: ko://github.com/awslabs/kit/cmd/controller
+        image: ko://github.com/awslabs/kubernetes-iteration-toolkit/operator/cmd/operator
```
ellistarn, line 127: I'd name this file controller.yaml or something. Or perhaps config/controller/deployment.yaml, config/controller/rbac/yaml

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn, line 55: I'd probably model this as the following
```
type MasterSpec struct {
  Scheduler Scheduler
  APIServer APIServer
}

type Scheduler struct {
  corev1.Deployment
  // Additional high level params go here, if any
}

type APIServer struct {
  corev1.Deployment
  // Additional high level params go here, if any
}
```

I'm also curious if it makes more sense to define this as a deployment spec or as a pod spec. I know that we may want to control replicas, so perhaps that's useful. I know ETCD will likely be a statefulset, rather than a deployment. I'm interested to see how this model progresses (i.e. being specific that we are running the Deployment abstraction and not the Pod abstraction).

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn: Huge +1. Let's use `corev1.Deployment`

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,39 @@ type ControlPlaneList struct {
...
 }
 
 // Instances denotes how the infrastructure of a particular components looks
 // like, if a user wants to use a specific AMI ID, they can provide this in the
 // Instances for the corresponding component.
 type Instances struct {
-	AMI   string `json:"ami,omitempty"`
-	Type  string `json:"type,omitempty"`
-	Count int    `json:"count,omitempty"`
+	AMI  string `json:"ami,omitempty"`
```
ellistarn, line 78: I'm not sure AMI is enough. You'll likely need to control user data as well. If we're using Karpenter, can we do this with NodeSelectors to customer LaunchTemplates (w/ custom ami)? That will allow us to drive everything via the corev1.DeploymentSpec

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn: Minimally, this buys us the ability to control the RolloutStrategy.

.github/workflows/build.yaml
prateekgogia: >but I consider the entire software installation (a set of controllers and webhooks) to together be an operator
Isn't this a complete installation (with set of controllers and webhooks), like whatever we need to create clusters will part of this binary?
The confusing part is, we have this project named as kit/operator and its binary is named controller?

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
prateekgogia: yes will be adding them in the next iterations.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
prateekgogia: The thing with using `v1.Deployment` object is it massively changes the user experience. 
This is how a user will create a default cluster-
```
apiVersion: kit.k8s.amazonaws.com/v1alpha1
kind: ControlPlane
metadata:
  name: example # Desired Cluster Name
spec: {}
```

Now if a user wants to change image name for api server and scheduler, they will need to do something like-
```
apiVersion: kit.k8s.amazonaws.com/v1alpha1
kind: ControlPlane
metadata:
  name: example # Desired Cluster Name
spec: 
  items:
  - master:
     scheduler:
       spec:
         template:
            spec:
              containers:
                 image: scheduler-image-name
     apiServer:
       spec:
         template:
            spec:
              containers:
                 image: apiserver-image-name
```
IMO, as a user I need to check the syntax for v1.Deployment everytime I need to make a small edit, at this point I might just go and edit the YAML spec in management cluster.

This field is implemented as a Kubernetes deployment object in the operator so we will be getting the rolling updates. Is there any functionality from the `v1.Deployment` spec that we need to expose to the user that we can't provide from `v1.PodSpec`?

hakuna-matatah: Speaking of rolling updates( taking a step back here), I was wondering do we actually need to update the spec of a test cluster that's already created ? I feel like that's more of an overkill to me than creating a new test cluster with the spec we wanted ? In my mind, I don't see  great benefits of updating a test cluster. Could you let me know what use-case I'm missing here  for megaxl or any test cases that we have today ?

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn: We will likely also need a ...Deployment to enable arbitrary configuration. Though I suppose this could be customized outside of the Cluster CR.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn: Use Case: test upgrading a cluster under load.

ellistarn: @prateekgogia, I see your argument. We could potentially go further and do a ContainerSpec instead of a podspec, but I think we may lose useful bits like node selectors. I'm supportive of a more customized DeploymentSpec (e.g. PodSpec and Replicas)

.github/workflows/build.yaml
ellistarn: We will likely have two binaries: controller and webhook. Perhaps even more binaries for different things. These are all pieces of the "operator"

ellistarn: e.g. Karpenter is a kubernetes operator with a controller and a webhook.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
ellistarn: What do you need custom AMI for in the control plane, any examples?

For any example you provide, I can think of a use case for user data. If you're going with custom AMI support, I think you need custom LT support.

ellistarn: We could potentially folow the same model as elsewhere and default the userdata to whatever the EKS Control Plane AMI expects.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
hakuna-matatah: Cool, that could be controlled through `Replicas` count on Deployment itself (which makes sense). But is there a benefit in updating and monitoring `PodSpec` per se which is going to be crux of the controller logic IIUC ?

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
prateekgogia: We can always add this later if needed. For now I am just keeping this limited.

operator/hack/codegen.sh
```diff
@@ -13,3 +13,10 @@ controller-gen \
 mv config/kit.k8s.amazonaws.com_controlplanes.yaml config/control-plane-crd.yaml
 # CRDs don't currently jive with VolatileTime, which has an Any type.
 perl -pi -e 's/Any/string/g' config/control-plane-crd.yaml
+
+# Kubectl apply fails if the annotations is too long with error -
+# The CustomResourceDefinition "controlplanes.kit.k8s.amazonaws.com" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
+# More info: https://stackoverflow.com/a/62409266
+yq eval 'del(.. | select(has("description")).description)' -i config/control-plane-crd.yaml
```
hakuna-matatah, line 20: haha

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,47 @@ type ControlPlaneList struct {
...
+	Type string `json:"type,omitempty"`
+}
+
+func (c *ControlPlane) ClusterName() string {
+	return c.Name
```
hakuna-matatah, line 83: do we have Name as part of ControlPlane spec ? I don't remember seeing it unless I'm misreading this.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,47 @@ type ControlPlaneList struct {
...
+}
+
+func (c *ControlPlane) NamespaceName() string {
+	return c.Namespace
```
hakuna-matatah, line 87: do we have Namespace as part of ControlPlane spec ( i remember seeing meta ? I don't remember seeing it unless I'm misreading this.

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
```diff
@@ -41,47 +42,47 @@ type ControlPlaneList struct {
...
 }
 
-// Config provides a generic way to pass in args and images to master and etcd
+// Component provides a generic way to pass in args and images to master and etcd
 // components. If a user wants to change the QPS they need to provide the
 // following flag with the desired value -`kube-api-qps:100` in the args.
-type Config struct {
-	Args  map[string]string `json:"args,omitempty"`
-	Image string            `json:"image,omitempty"`
+type Component struct {
```
hakuna-matatah, line 69: This sounds way better 💯

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
prateekgogia: Yes, its part of the `metadata` field

operator/pkg/apis/infrastructure/v1alpha1/controlplane.go
prateekgogia: Yes, same as above.

.gitignore
```diff
@@ -0,0 +1 @@
+operator/bin
```
ellistarn, line 1: Why are you ever even generating binaries like this?

operator/cmd/controller/main.go
```diff
@@ -56,5 +56,7 @@ func main() {
 
 	_ = awsprovider.NewSession()
 	err := manager.RegisterWebhooks().RegisterControllers().Start(controllerruntime.SetupSignalHandler())
-	log.PanicIfError(err, "Unable to start manager")
```
ellistarn, line 59: +1

operator/pkg/awsprovider/session.go
```diff
@@ -24,8 +24,10 @@ import (
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/awslabs/karpenter/pkg/utils/log"
-	"github.com/awslabs/karpenter/pkg/utils/project"
+)
+
+const (
+	userAgent = "operator-kit.sh"
```
ellistarn, line 30: "kubernetes-iteration-toolkit/operator"

ellistarn: I don't expect we'll ever use a .sh domain.

ellistarn (dismissed):
A couple optional nits

hakuna-matatah (approved):

operator/pkg/awsprovider/session.go
prateekgogia: Thanks, updated.

ellistarn (approved):

hakuna-matatah (approved):

PR #10, opened 2021-07-09
TestExecution UI for KIT

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
TestExecution UI for KIT which leverages AWSLoadbalancer to create a LB in front of Dashboard service ; Cleaned up some addons; Making flux to point to `testbed` branch for now so it's much easier for testing changes before creating a PR.


testbed/addons/karpenter/construct.ts
```diff
@@ -4,24 +4,25 @@ import iam = require('@aws-cdk/aws-iam');
 
 export interface KarpenterProps {
     cluster: eks.Cluster;
-    namespace?: string;
```
ellistarn, line 7: I actually thought it was nice that we could configure this.

testbed/addons/karpenter/construct.ts
```diff
@@ -4,24 +4,25 @@ import iam = require('@aws-cdk/aws-iam');
...
+    private namespace: string;
     constructor(scope: cdk.Construct, id: string, props: KarpenterProps) {
         super(scope, id);
-
-        const ns = props.cluster.addManifest('namespace', {
-            apiVersion: 'v1',
-            kind: 'Namespace',
-            metadata: {
-                name: props.namespace ?? "karpenter"
+        this.namespace = "karpenter"
```
ellistarn, line 13: instead, consider props.namespace = props.namespace ?? "karpenter", and then just use that existing field. There might even be an operator `props.namespace ?= "karpenter"`

testbed/addons/karpenter/construct.ts
```diff
@@ -4,24 +4,25 @@ import iam = require('@aws-cdk/aws-iam');
...
+        const ns = props.cluster.addManifest('namespace',
```
ellistarn, line 14: mmmm this makes sense :). I wonder if there's an idempotency challenge if the user wanted to pass in a namespace. Maybe best to crush all of this complexity and just make it not configurable.

testbed/addons/karpenter/construct.ts
```diff
@@ -4,24 +4,25 @@ import iam = require('@aws-cdk/aws-iam');
 
 export interface KarpenterProps {
     cluster: eks.Cluster;
-    namespace?: string;
 }
 
 export class Karpenter extends cdk.Construct {
+    private namespace: string;
```
ellistarn, line 10: Depending on your impl (props vs local), consider using a const instead of a private variable, since this value isn't used within member functions of this class. It's only used in the ctor.

testbed/addons/karpenter/construct.ts
```diff
@@ -47,7 +48,7 @@ export class Karpenter extends cdk.Construct {
             ]
         }),
             {
-                username: `system:node:{{EC2PrivateDNSName}}`,
+                username: 'system:node:{{EC2PrivateDNSName}}',
```
ellistarn, line 51: Nice catch! Typescript ``` syntax is for string interpolation like `concatenate-to-${this.namespace}`

testbed/addons/tekton-dashboard/release.yaml
```diff
@@ -0,0 +1,492 @@
+apiVersion: apiextensions.k8s.io/v1
```
ellistarn, line 1: Can you add a comment or a script somewhere for how to programatically download or generate this? I don't have a problem checking it in if it's not available via helm somewhere.

testbed/bin/testbed.ts
```diff
@@ -9,5 +9,5 @@ new TestbedStack(app, 'TestbedStack', {
         region: process.env.CDK_DEFAULT_REGION
     },
     repositories: [
-        { name: "testbed", url: "https://github.com/awslabs/kubernetes-iteration-toolkit", path: "./testbed/addons" }]
+        { name: "testbed", url: "https://github.com/awslabs/kubernetes-iteration-toolkit", path: "./testbed/addons", branch: "testbed" }]
```
ellistarn, line 12: Hmm -- why do we want to store this on a branch called "testbed"?

testbed/addons/karpenter/construct.ts
hakuna-matatah: Spoke offline, leaning to convention here than configurability. Same for following other comments on this PR where applicable.

testbed/bin/testbed.ts
hakuna-matatah: Like mentioned in the description, it's easier for me to test end to end (using flux)  before creating a PR. Once it's fully ready for users to use it, I will point it to HEAD on main.

testbed/addons/tekton-dashboard/release.yaml
hakuna-matatah: Spoke offline. Earlier it was programmatically installed through flux from tekton dashboard repo(https://github.com/awslabs/kubernetes-iteration-toolkit/pull/10/files#diff-12f8aaa63c2e28fe96b5d47b3651cfc21103ea346a1a91594da07aab395b5cb8L4) . As I mentioned in description, we are customizing this yaml for our testbed purposes to have a UI, hence checking in the yaml.

hakuna-matatah: I'll do CDK deploy locally without this change as workaround.

testbed/bin/testbed.ts
hakuna-matatah: I will do a CDK deploy of this change locally as workaround.

ellistarn (approved):

PR #18, opened 2021-08-03
Add certs for master components

prateekgogia:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/operator/pkg/controllers/etcd/etcd.go
+++ b/operator/pkg/controllers/etcd/etcd.go
@@ -24,6 +24,7 @@ import (
 	"github.com/awslabs/kit/operator/pkg/apis/infrastructure/v1alpha1"
 	pkiutil "github.com/awslabs/kit/operator/pkg/pki"
 	"github.com/awslabs/kit/operator/pkg/utils/patch"
+	"github.com/awslabs/kit/operator/pkg/utils/secrets"
 
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


operator/pkg/controllers/etcd/etcd.go
```diff
@@ -115,59 +108,17 @@ func (e *Provider) createService(ctx context.Context, controlPlane *v1alpha1.Con
...
-		return fmt.Errorf("creating root CA for name %v, %w", controlPlane.ClusterName(), err)
-	}
-	// create etcd server, peer, healthcheck and etcdAPIClient certs and key
-	for _, certConfig := range certListFor(controlPlane, caCert, caKey) {
-		if _, _, err := e.createSecret(ctx, controlPlane, certConfig); err != nil {
-			return fmt.Errorf("creating certs and key name %s, %w,", certConfig.Name, err)
-		}
+func (p *Provider) createETCDSecrets(ctx context.Context, controlPlane *v1alpha1.ControlPlane) error {
+	// create the root CA, certs and key for etcd
+	if err := secrets.WithRootCAName(p.kubeClient,
```
ellistarn, line 113: I'd avoid passing dependencies like p.kubeClient as a parameter and instead inject it at object initialization time as a private member.

operator/pkg/controllers/master/master.go
```diff
@@ -0,0 +1,152 @@
...
+	kubeClient client.Client
+}
+
+func New(kubeclient client.Client) *Provider {
+	return &Provider{kubeClient: kubeclient}
+}
+
+func (p *Provider) Reconcile(ctx context.Context, controlPlane *v1alpha1.ControlPlane) (err error) {
+
+	// TODO Create a service for master NLB
```
ellistarn, line 47: I'd avoid referencing details like NLB, since it's the implementation of the service.

operator/pkg/controllers/master/master.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	masterCACommonName     = "kubernetes"
+	frontProxyCACommonName = "front-proxy-ca"
+)
+
+type Provider struct {
```
ellistarn, line 37: I might call this a controller, rather than a Provider

operator/pkg/controllers/master/master.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := p.createMasterSecrets(ctx, controlPlane); err != nil {
+		return err
+	}
+	// TODO Create and apply objects for kube apiserver, KCM and scheduler
+	return nil
+}
+
+// createMasterSecrets creates the kubernetes secrets containing all the certs
+// and key required to run master API server
+func (p *Provider) createMasterSecrets(ctx context.Context, controlPlane *v1alpha1.ControlPlane) error {
```
ellistarn, line 57: You already know it's related to master by the package name. Consider just "createSecrets"

operator/pkg/utils/secrets/secrets.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// create creates a v1.Secret object that contains the cert and key and is
+// stored in Kubernetes cluster. If the secret object is found in the cluster,
+// it reuses the existing the existing cert and key if it is valid.
+func (c *CA) create(ctx context.Context, controlPlane *v1alpha1.ControlPlane, req *Request) (cert, key []byte, err error) {
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      req.Name,
+			Namespace: controlPlane.NamespaceName(),
```
ellistarn, line 99: Does it make sense to put namespace as part of the request rather than getting it from the controlplane?

operator/pkg/utils/secrets/secrets.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		},
+	})
+	return
+}
+
+// create creates a v1.Secret object that contains the cert and key and is
+// stored in Kubernetes cluster. If the secret object is found in the cluster,
+// it reuses the existing the existing cert and key if it is valid.
+func (c *CA) create(ctx context.Context, controlPlane *v1alpha1.ControlPlane, req *Request) (cert, key []byte, err error) {
```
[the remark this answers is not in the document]
prateekgogia: The challenge with this is duplicate code in the higher level controller. Controller needs to be aware of the caCert and caKey as well and then pass in the right CA when the cert/key is created to sign.
I wanted to abstract this away from controller. As in, controller should be able to make calls like `secrets.RootCA().Create(req...)`

Happy to discuss more on a call.

operator/pkg/utils/secrets/secrets.go
prateekgogia: It does, but it gets confusing as we can't avoid passing in `controlPlane` so why not just get namespace from `controlPlane` object.

ellistarn (approved):

PR #20, opened 2021-08-04
Adds control plane service object  for cluster endpoint

prateekgogia:
*Issue #, if available:*

*Description of changes:*
Create a Kubernetes service object of type Loadbalancer for cluster endpoint 

### DIFF
--- a/operator/pkg/controllers/controller.go
+++ b/operator/pkg/controllers/controller.go
@@ -21,7 +21,7 @@ import (
 
 	"github.com/awslabs/kit/operator/pkg/apis/infrastructure/v1alpha1"
 	"github.com/awslabs/kit/operator/pkg/errors"
-	"github.com/awslabs/kit/operator/pkg/status"
+	"github.com/awslabs/kit/operator/pkg/result"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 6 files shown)


operator/pkg/controllers/master/master.go
```diff
@@ -69,28 +77,76 @@ func (c *Controller) createSecrets(ctx context.Context, controlPlane *v1alpha1.C
 	return nil
 }
 
-func certListFor(controlPlane *v1alpha1.ControlPlane) []*secrets.Request {
+func (c *Controller) getControlPlaneHostname(ctx context.Context, controlPlane *v1alpha1.ControlPlane) (string, error) {
```
ellistarn, line 80: Seems odd to me that the "get" method would "create" internally. 

I might model this like:
```
service, err := c.client.ensure(ctx, c.ServiceFor(controlPlane))
hostname := service.Status.Loadbalancer.Ingress
if  == nil {
  return result.Wait()
}

c.SecretsFor(hostname)
...
```

operator/pkg/controllers/master/master.go
```diff
@@ -69,28 +77,76 @@ func (c *Controller) createSecrets(ctx context.Context, controlPlane *v1alpha1.C
...
+	hostname, err := c.createService(ctx, controlPlane)
+	if err != nil {
+		return "", err
+	}
+	if hostname == "" {
+		return "", fmt.Errorf("waiting for control plane hostname, %w", errors.WaitingForSubResources)
```
ellistarn, line 86: In the long run, let's model this as a reconcile.result instead of munging it into a typed error.

ellistarn (approved):

operator/pkg/controllers/controller.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/awslabs/kit/operator/pkg/apis/infrastructure/v1alpha1"
 	"github.com/awslabs/kit/operator/pkg/errors"
-	"github.com/awslabs/kit/operator/pkg/status"
+	"github.com/awslabs/kit/operator/pkg/result"
```
JacobGabrielson, line 24: I'd suggest renaming this pkg to `results` maybe? It's that that `result` is such a common word to actually use (especially in Go), even if you don't have actual naming conflicts, every time someone sees `result` they're going to thing it's a local variable (containing a result).

operator/pkg/controllers/controller.go
```diff
@@ -48,40 +48,43 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
...
+		}
+		return *result.Failed, reconcileErr
 	}
-	return result, nil
+	return reconcileResult, nil
 }
 
 func (c *GenericController) reconcile(ctx context.Context, resource Object, persisted runtime.Object) (reconcile.Result, error) {
-	var result *reconcile.Result
+	var reconcileResult *reconcile.Result
```
JacobGabrielson, line 71: See earlier comment - I think having to do this renaming is probably a sign that that `result` pkg's name should change instead.

operator/pkg/controllers/controller.go
prateekgogia: Thanks, yeah I will address this as part of the next PR I am working on.

PR #22, opened 2021-08-10
Control plane kubeconfigs and secret refactoring

prateekgogia:
*Issue #, if available:*

*Description of changes:*
- Control plane kubeconfigs and secret refactoring
- Rename `result` pkg to `results`


operator/pkg/controllers/etcd/etcd.go
```diff
@@ -20,181 +20,77 @@ import (
...
-	if err := c.createSecrets(ctx, controlPlane); err != nil {
-		return err
-	}
-	// create etcd stateful set
-	if err := c.createStatefulset(ctx, controlPlane); err != nil {
-		return err
-	}
-	return nil
+func New(kubeclient *kubeprovider.Client) *Controller {
+	return &Controller{kubeClient: kubeclient, certificates: common.New(kubeclient)}
```
ellistarn, line 46: common.New feels a bit off to me. Aren't all of our controllers using the same kubeprovider.Client object?

operator/pkg/controllers/etcd/etcd.go
```diff
@@ -20,181 +20,77 @@ import (
...
+	for _, reconcile := range []reconciler{
+		c.reconcileService,
+		c.reconcileSecrets,
+		c.reconcileStatefulSet,
+	} {
+		if err := reconcile(ctx, controlPlane); err != nil {
+			return err
+		}
 	}
+	zap.S().Infof("[%v] etcd reconciled", controlPlane.ClusterName())
```
ellistarn, line 59: These types of logs are gonna get a bit noisy. Consider only logging when you take explicit action.

operator/pkg/controllers/etcd/etcd.go
```diff
@@ -20,181 +20,77 @@ import (
...
-	return fmt.Sprintf("etcd-server-ssl-%s", clusterName)
-}
-
-func etcdClientPortNameFor(clusterName string) string {
-	return fmt.Sprintf("etcd-client-ssl-%s", clusterName)
-}
-
-func etcdServiceNameFor(clusterName string) string {
-	return fmt.Sprintf("%s-etcd", clusterName)
+	return c.certificates.ReconcileFor(ctx, secretTreeMap, cp)
```
ellistarn, line 72: More idiomatic to invert these, imo. Reconcile(ctx, controlplane, secrets)

operator/pkg/controllers/etcd/etcd.go
```diff
@@ -20,181 +20,77 @@ import (
...
-		"app": etcdServiceNameFor(clusterName),
-	}
-}
-
-func certListFor(controlPlane *v1alpha1.ControlPlane) []*secrets.Request {
-	return []*secrets.Request{
-		etcdServerCertConfig(controlPlane),
-		etcdPeerCertConfig(controlPlane),
+func rootCACertConfig(nn types.NamespacedName) *secrets.Request {
+	return &secrets.Request{
```
ellistarn, line 88: If you wanted to be really fancy, secrets.Request could take a namespacedname instead of separating the fields, so you could do

```
secrets.Request { NamespacedName: nn, ... } 
```

operator/pkg/controllers/etcd/etcd.go
```diff
@@ -208,18 +104,17 @@ The last two entries are added for every pod in the cluster
...
-				CommonName:   "etcd",
-				Organization: []string{"kubernetes"},
-				AltNames: certutil.AltNames{
-					DNSNames: append(etcdPodAndHostnames(controlPlane),
-						etcdSvcFQDN(controlPlane.ClusterName(), controlPlane.Namespace),
-						"localhost"),
-					IPs: []net.IP{net.IPv4(127, 0, 0, 1)},
-				},
+		Name:      serverSecretNameFor(controlPlane.ClusterName()),
+		Namespace: controlPlane.NamespaceName(),
```
ellistarn, line 108: I did a double take when reading this, since I read NamespacedName instead of NamespaceName. I might just use Namespace directly.

operator/pkg/controllers/etcd/statefulset.go
```diff
@@ -0,0 +1,64 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (c *Controller) reconcileStatefulSet(ctx context.Context, controlPlane *v1alpha1.ControlPlane) error {
+	// Generate the default pod spec for the given control plane, if user has
+	// provided custom config for the etcd pod spec, patch this user
+	// provided config to the default spec
+	etcdSpec, err := patch.PodSpec(podSpecFor(controlPlane), controlPlane.Spec.Etcd.Spec)
```
ellistarn, line 35: Given that we control the types on both pod specs, can we ensure that you can't runtime errors of the sort on line 37? Be squashing this error case, this simplifies a bunch of things. For one, it allows you do inline everything on line 39, including the full pod spec, which could be decorated inline. 
```
spec: patch.PodSpec(&PodSpec{ ... }, controlplane.Spec.EtcdSpec)
```
I guess it's not perfectly clean, since it's a bit awkward to have the `controlplane.Spec.EtcdSpec` dangling, but it doesn't make sense to put it as the first parameter w/ the patch semantic.

operator/pkg/controllers/etcd/service.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (c *Controller) reconcileService(ctx context.Context, controlPlane *v1alpha1.ControlPlane) error {
+	return c.kubeClient.Ensure(ctx, object.WithOwner(controlPlane, &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceNameFor(controlPlane.ClusterName()),
+			Namespace: controlPlane.NamespaceName(),
+			Labels:    labelFor(controlPlane.ClusterName()),
+		},
+		Spec: v1.ServiceSpec{
+			ClusterIP: v1.ClusterIPNone,
```
ellistarn, line 37: ClusterIPNone seems wrong here. Can we leave this empty?

operator/pkg/controllers/master/kubeconfigs.go
```diff
@@ -0,0 +1,153 @@
...
+	} {
+		if err := c.reconcileConfigFor(ctx, request, controlPlane); err != nil {
+			return err
+		}
+	}
+	zap.S().Debugf("[%v] Kube configs reconciled", controlPlane.ClusterName())
+	return nil
+}
+
+func (c *Controller) reconcileConfigFor(ctx context.Context, request *secrets.Request, controlPlane *v1alpha1.ControlPlane) error {
```
ellistarn, line 49: Consider inverting controlplane and the secret request. My vague sense of why is that the control plane is the higher order object/concept, so it should come first.

operator/pkg/controllers/master/kubeconfigs.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil
+	}
+	// TODO validate the existing config in the secret
+	return err
+}
+
+func kubeConfigFor(userName, clusterName, endpoint string, caSecret *v1.Secret, cert, key []byte) *clientcmdapi.Config {
+	contextName := fmt.Sprintf("%s@%s", userName, clusterName)
+	caCert, _ := secrets.Parse(caSecret)
+	return &clientcmdapi.Config{
```
ellistarn, line 89: Super cool you found this library.

operator/pkg/controllers/master/master.go
```diff
@@ -16,206 +16,34 @@ package master
...
-
-func labelsFor(clusterName string) map[string]string {
-	return map[string]string{
-		"app": serviceNameFor(clusterName),
-	}
-}
-
-func apiserverPortName(clusterName string) string {
-	return fmt.Sprintf("%s-port", serviceNameFor(clusterName))
+	zap.S().Infof("[%v] control plane reconciled", controlPlane.ClusterName())
```
ellistarn, line 47: same here.

operator/pkg/kubeprovider/client.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/awslabs/kit/operator/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Client struct {
+	client.Client
+}
+
+func New(client client.Client) *Client {
+	return &Client{Client: client}
```
ellistarn, line 30: Have you seen the ordered parameters syntax before?

`return &Client{client}`

operator/pkg/utils/common/certificates.go
```diff
@@ -0,0 +1,92 @@
...
+type CertificatesProvider struct {
+	kubeClient *kubeprovider.Client
+}
+
+// CertTree contains root CA as a the key and all the certificates signed by
+// this root CA are added as the value for this map
+type CertTree map[*secrets.Request][]*secrets.Request
+
+func New(kubeClient *kubeprovider.Client) *CertificatesProvider {
+	return &CertificatesProvider{kubeClient: kubeClient}
```
ellistarn, line 40: likewise

operator/pkg/utils/common/certificates.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/awslabs/kit/operator/pkg/errors"
+	"github.com/awslabs/kit/operator/pkg/kubeprovider"
+	"github.com/awslabs/kit/operator/pkg/utils/object"
+	"github.com/awslabs/kit/operator/pkg/utils/secrets"
+	"go.uber.org/zap"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type CertificatesProvider struct {
```
ellistarn, line 31: Consider reorienting this thing to 
```
utils/certificates/reconciler.go

certificates.NewReconciler(...)
```

It might even make sense for this to share the same package of your utils/secret package. Even

```
certificates.SecretRequest
```

operator/pkg/utils/object/object.go
```diff
@@ -0,0 +1,35 @@
...
+	obj.SetOwnerReferences([]metav1.OwnerReference{{
+		APIVersion: owner.GetObjectKind().GroupVersionKind().Version,
+		Name:       owner.GetName(),
+		Kind:       owner.GetObjectKind().GroupVersionKind().Kind,
+		UID:        owner.GetUID(),
+	}})
+	return obj
+}
+
+func NamespacedName(name, namespace string) types.NamespacedName {
```
ellistarn, line 33: This helper smells a bit funny to me. Is it worth using over types.NamespacedName?

ellistarn (approved):
Very nicely done! Committable as is, but some suggestions for further refinements if you're interested.

operator/go.mod
```diff
@@ -5,9 +5,11 @@ go 1.16
 require (
 	github.com/aws/aws-sdk-go v1.38.62
 	github.com/awslabs/karpenter v0.2.8
+	github.com/ghodss/yaml v1.0.0
```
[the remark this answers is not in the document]
prateekgogia: Yes, `go mod tidy` added this, I double checked its not needed. It was coming from some test files I have in my local repo. Removing this.

operator/pkg/controllers/etcd/etcd.go
prateekgogia: Will revisit this later.

operator/pkg/controllers/etcd/statefulset.go
prateekgogia: Keeping this as is for now, I tried changing it but not sure, if we can avoid an error while patching.

operator/pkg/controllers/etcd/service.go
prateekgogia: Need to explicitly set this to `None`, to make this a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services).

operator/pkg/utils/object/object.go
prateekgogia: I am still debating if it makes it cleaner to have this `func(nn types.NamespacedName)` or should I revert back to `func(name, ns string)`
Will revisit this.

.gitignore
```diff
@@ -1 +1,2 @@
 operator/bin
+.vscode/
```
ellistarn, line 2: leave this one out

ellistarn (approved):

PR #29, opened 2021-08-20
Deployments for Master components and unit-tests

prateekgogia:
*Description of changes:*
- Adds deployments for Master components(APIServer, KCM, Scheduler)
- Adds unit tests to check all components get created


.github/workflows/image.yaml
```diff
@@ -22,8 +22,9 @@ jobs:
           go-version: 1.16
       - name: Build & Test
         run: |
-          go build -v ./cmd/controller
-          go test ./...
+          make toolchain
+          echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
+          make build && make test
```
ellistarn, line 27: I think you can say: `make toolchain build test`

operator/pkg/controllers/controlplane/suite_test.go
```diff
@@ -0,0 +1,117 @@
...
+		Context("Objects", func() {
+			It("should create all desired objects in the same namespace", func() {
+				// Create the control plane cluster
+				ExpectCreated(kubeClient, controlPlane)
+				// Reconcile the controller for the given control plane
+				genController := &controllers.GenericController{Controller: controller, Client: kubeClient}
+				ExpectReconcile(context.Background(), genController, client.ObjectKeyFromObject(controlPlane))
+				// check master and etcd services are created
+				ExpectServiceExists(kubeClient, etcd.ServiceNameFor(controlPlane.Name), controlPlane.Namespace)
+				ExpectServiceExists(kubeClient, master.ServiceNameFor(controlPlane.Name), controlPlane.Namespace)
```
ellistarn, line 86: Can you write a generic function "ExpectExists"?

operator/pkg/controllers/etcd/secret.go
```diff
@@ -0,0 +1,153 @@
...
+		Namespace: controlPlane.Namespace,
+		Type:      secrets.KeyWithSignedCert,
+		Config: &certutil.Config{
+			Usages:       []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth, x509.ExtKeyUsageClientAuth},
+			CommonName:   "etcd",
+			Organization: []string{"kubernetes"},
+			AltNames: certutil.AltNames{
+				DNSNames: append(etcdPodAndHostnames(controlPlane),
+					SvcFQDN(controlPlane.ClusterName(), controlPlane.Namespace),
+					"localhost"),
```
ellistarn, line 92: nit: I'd just keep this one on line 91

operator/pkg/controllers/etcd/secret.go
```diff
@@ -0,0 +1,153 @@
...
```
ellistarn, line 119: likewise

operator/pkg/controllers/master/apiserverdeployment.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/awslabs/kit/operator/pkg/controllers/etcd"
+	"github.com/awslabs/kit/operator/pkg/utils/object"
+	"github.com/awslabs/kit/operator/pkg/utils/patch"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TODO move to master.go
```
ellistarn, line 32: I'd just do this right now, rather than TODOing it.

operator/pkg/controllers/master/apiserverdeployment.go
```diff
@@ -0,0 +1,324 @@
...
+func nodeSelector(clusterName string) map[string]string {
+	return map[string]string{
+		"node":        "control-plane",
```
ellistarn, line 35: Why do you need this label?

operator/pkg/controllers/master/apiserverdeployment.go
```diff
@@ -0,0 +1,324 @@
...
+		"clustername": clusterName,
```
ellistarn, line 36: consider something like `kit.kubernetes.io/control-plane-name: name`

operator/pkg/controllers/master/apiserverdeployment.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+}
+
+const (
+	apiserverImage        = "public.ecr.aws/eks-distro/kubernetes/kube-apiserver:v1.20.7-eks-1-20-4"
```
ellistarn, line 41: We should unify all these somewhere. Maybe like a distribution provider? distributionProvider.getEtcdImage(kubernetesVersion)

operator/pkg/controllers/master/apiserverdeployment.go
```diff
@@ -0,0 +1,324 @@
...
+
+func APIServerDeploymentName(clusterName string) string {
+	return fmt.Sprintf("%s-apiserver", clusterName)
+}
+
+func apiServerLabels(clustername string) map[string]string {
+	return patch.UnionStringMaps(labelsFor(clustername), apiserverComponent)
+}
+
+var apiserverComponent = map[string]string{"component": "apiserver"}
```
ellistarn, line 82: just inline, imo

operator/pkg/controllers/master/kcmdeployment.go
```diff
@@ -0,0 +1,211 @@
+/*
```
ellistarn, line 1: nit: kube-controller-manager.go

ellistarn: Make folders if you want to split deployment/service/etc into separate files.

/controllers/master/kube-controller-manager/deployment.go

operator/pkg/controllers/master/kcmdeployment.go
```diff
@@ -0,0 +1,211 @@
...
+
+func KCMDeploymentName(clusterName string) string {
+	return fmt.Sprintf("%s-controller-manager", clusterName)
+}
+
+func kcmLabels(clustername string) map[string]string {
+	return patch.UnionStringMaps(labelsFor(clustername), kcmComponent)
+}
+
+var kcmComponent = map[string]string{"component": "kube-controller-manager"}
```
ellistarn, line 68: likewise.

operator/pkg/controllers/master/schedulerdeployment.go
```diff
@@ -0,0 +1,148 @@
+/*
```
ellistarn, line 1: same. scheduler -> kube-scheduler

operator/pkg/controllers/master/schedulerdeployment.go
```diff
@@ -0,0 +1,148 @@
...
+
+func SchedulerDeploymentName(clusterName string) string {
+	return fmt.Sprintf("%s-scheduler", clusterName)
+}
+
+func schedulerLabels(clustername string) map[string]string {
+	return patch.UnionStringMaps(labelsFor(clustername), schedulerComponent)
+}
+
+var schedulerComponent = map[string]string{"component": "kube-scheduler"}
```
ellistarn, line 68: likewise

operator/pkg/test/environment/fakekubeclient.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type FakeKubeClient struct {
```
ellistarn, line 25: Trying to understand why you need this at all. What's the special bits about the string suffix? Maybe we can discuss offline?

operator/pkg/test/expectations/expectations.go
```diff
@@ -0,0 +1,115 @@
...
+	return deployment
+}
+
+func ExpectStatefulSetExists(c client.Client, name string, namespace string) *appsv1.StatefulSet {
+	set := &appsv1.StatefulSet{}
+	Expect(c.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, set)).To(Succeed())
+	return set
+}
+
+func ExpectCleanedUP(c client.Client) {
```
ellistarn, line 74: UP -> Up

ellistarn (commented):
This PR was a delight to review. Nicely done. A few nits here and there.

.github/workflows/image.yaml
prateekgogia: Nice! Thanks!

operator/pkg/controllers/controlplane/suite_test.go
prateekgogia: Do you mean like a wrapper that calls these sub-functions or call `ExpectExists` here and pass the object type as an argument to this func?

operator/pkg/controllers/master/apiserverdeployment.go
prateekgogia: Need to separate control-plane and etcd nodes.

operator/pkg/controllers/master/apiserverdeployment.go
prateekgogia: Will revisit this as this needs more work. I need to figure out how to accept Kubernetes version from user, so will do this as part of that task.

operator/pkg/test/environment/fakekubeclient.go
prateekgogia: Master has a dependency on control plane service to be created and loadbalancer hostname to be populated before it can generate certs. In tests, I need to make sure hostname gets populated.
Let's discuss this more offline.

operator/pkg/controllers/controlplane/suite_test.go
ellistarn: Your function doesn't rely on any of the concrete type information, so you can use a runtimeobject. Karpenter has some other helpers that *does* rely on the type information because it returns the object which can be nice for chaining
```
func ExpectPodExists(c client.Client, name string, namespace string) *v1.Pod {
	pod := &v1.Pod{}
	Expect(c.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, pod)).To(Succeed())
	return pod
}

func ExpectNodeExists(c client.Client, name string) *v1.Node {
	node := &v1.Node{}
	Expect(c.Get(context.Background(), client.ObjectKey{Name: name}, node)).To(Succeed())
	return node
}
```
In your case, since you're not asserting anything about the service, I think the former ExpectExists(runtime.Object) makes the most sense.

operator/pkg/controllers/master/apiserverdeployment.go
ellistarn: Best practice for node labels is to qualify with a domain so that readers can know where they come from. e.g. `kit.kubernetes.io/node: control-plane`.

ellistarn (approved):

PR #31, opened 2021-08-24
Testworkflows - upload results to s3 bucket

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*

Testworkflows - upload results to s3 bucket


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -73,3 +73,7 @@ spec:
     env:
     - name: ENABLE_EXEC_SERVICE
       value: "false"
+  - name: upload-results
+    image: amazon/aws-cli
+    workingDir: $(workspaces.results.path)
+    command: ["aws s3 cp . s3://$(params.results-bucket)/$(date +'%Y/%m/%d/%HH/%mm')  --recursive "]
```
ellistarn, line 79: Can we include seconds in this timestamp? Or perhaps a uuid? I can see this colliding otherwise.

tests/pipelines/eks/small/upstream-load.yaml
```diff
@@ -56,3 +56,5 @@ spec:
       params:
       - name: cluster-name
         value: "eks-small-upstream-load-$(context.pipelineRun.uid)"
+      - name: results-bucket
+        value: "$(date +'%Y/%m/%d/%HH/%mm)/$(context.pipelineRun.uid)"
```
ellistarn, line 60: Nice! The only downside of this is we now need to configure this in every one of our pipelines, which is kind of a bummer. It would be nice if the task could somehow be aware of the taskrun and use the ID there as part of the CL2 spec.

ellistarn (approved):

PR #38, opened 2021-08-31
Adds Dataplane support and core DNS add ons

prateekgogia:
*Description of changes:*
Still need to clean up and rename a few things
- Brings up an ASG with worker nodes for the guest cluster control plane
- Adds core DNS addon for the guest cluster control plane

### Todo

- [x] - Add authenticator daemonset to be deployed 
- [x] - Add CNI addon for the guest cluster control plane


operator/docs/examples/dataplane.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kit.k8s.sh/v1alpha1
+kind: DataPlane
+metadata:
+  name: example-nodes 
+spec: 
+  clusterName: example # Desired Cluster Name
```
ellistarn, line 6: Thoughts on making this fully decoupled from the custom resource and doing something like

```
replicas: 5 # instead of nodecount, node is implied by the fact that this is a dataplane.
cluster:
  name: ...
  endpoint: ...
```

ellistarn (approved):
demo ready :)

PR #46, opened 2021-10-15
Fixes Readme, adds dataplane spec to charts

prateekgogia:
*Issue #, if available:*

*Description of changes:*


operator/charts/kit-operator/templates/data-plane-crd.yaml
```diff
@@ -0,0 +1,93 @@
+
```
hakuna-matatah, line 1: oh! nice.  This crd def was totally missing earlier then  :D

PR #48, opened 2021-10-18
Mege flags fix and sample spec for flags

prateekgogia:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/operator/pkg/utils/patch/patch.go
+++ b/operator/pkg/utils/patch/patch.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/awslabs/kit/operator/pkg/utils/functional"
+	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


operator/pkg/utils/patch/patch.go
```diff
@@ -70,7 +70,7 @@ func parseArgsFor(podSpec *v1.PodSpec) map[string]string {
 	result := map[string]string{}
 	for _, arg := range podSpec.Containers[0].Args {
 		kv := strings.Split(arg, "=")
-		result[kv[0]] = result[kv[1]]
+		result[kv[0]] = kv[1]
```
hakuna-matatah, line 73: Nice! So earlier to this fix, changes to the flags from spec weren't working I assume ?

prateekgogia: yes, this broke the merge functionality.

PR #51, opened 2021-10-19
Bug fix ; Update heml repo;CP spec and DP spec;

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
- Bug fix ; Update heml repo;CP spec and DP spec;

### DIFF
--- a/tests/tasks/setup/kit-cluster/default.yaml
+++ b/tests/tasks/setup/kit-cluster/default.yaml
@@ -17,10 +17,27 @@ spec:
   - name: guest-cluster-region
     default: "us-west-2"
     description: The region where the kit cluster is created.
-  - name: control-plane-spec
-    description:  url to the controlplane spec.
-  - name: dataplane-spec
-    description: url to the dataplane spec.
+  - name: version
+    default: "1.21"
+    description:  kubernetes version.
+  - name: cp-instance-type
+    default: "m5.8xlarge"
+    description: control plane instance type.
+  - name: max-requests-inflight
+    default: "400"
+    description: maximum number of inflight read request that apiserver could allow
+  - name: max-mutating-requests
+    default: "200"
+    description: maximum number of mutating requests in flight that apiserver could allow
+  - name: kcm-qps
+    default: "20"
+    description: Kubernetes-Controller-Manager QPS setting
+  - name: scheduler-qps
+    default: "20"
+    description: Kubernetes-Scheduler QPS setting
+  - name: node-count
+    default: "1000"
+    description: desired node count for Dataplane.
   workspaces:
   - name: config
     description: |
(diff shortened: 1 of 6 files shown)


tests/triggers/kit/small/cp.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: kit.k8s.sh/v1alpha1
+kind: ControlPlane
+metadata:
+  name: kit-small # Desired Cluster Name
```
prateekgogia, line 4: What does small refer to here?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -35,9 +52,51 @@ spec:
       aws eks update-kubeconfig --name testbed --region $(params.host-cluster-region)
       kubectl config current-context
       #Provision a control plane for the guest cluster
-      kubectl create -f $(params.control-plane-spec)
+      cat > "$(workspaces.config.path)/cp.yaml" <<EOL
+      apiVersion: kit.k8s.sh/v1alpha1
+      kind: ControlPlane
+      metadata:
+        name: kit-small # Desired Cluster Name
```
prateekgogia, line 59: Cluster name should be a parameter

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -17,10 +17,27 @@ spec:
   - name: guest-cluster-region
     default: "us-west-2"
     description: The region where the kit cluster is created.
-  - name: control-plane-spec
-    description:  url to the controlplane spec.
-  - name: dataplane-spec
-    description: url to the dataplane spec.
+  - name: version
+    default: "1.21"
```
prateekgogia, line 21: Not clear to me why do we need defaults in two places, pipelines and here?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -35,9 +52,51 @@ spec:
...
+      kubectl create -f $(workspaces.config.path)/cp.yaml
+
+      cat > "$(workspaces.config.path)/dp.yaml" <<EOL
+      apiVersion: kit.k8s.sh/v1alpha1
+      kind: DataPlane
+      metadata:
+        name: $(params.cluster-name)-nodes 
+      spec: 
+      clusterName: $(params.cluster-name) # Desired Cluster Name
+      nodeCount: $(params.node-count)
```
prateekgogia, line 94: Indentation seems wrong.

prateekgogia (changes requested):

tests/tasks/setup/kit-cluster/default.yaml
prateekgogia: Do we want that, as its a lot of duplication and if we need to change default we have to change them at 2 locations. Not a blocker for now but we should reconsider this.

PR #59, opened 2021-11-02
Fix charts to not create SA, testbed names

prateekgogia:
*Description of changes:*
- Fix create service account issue when deploying testbed, SA already exists and charts should skip creating it.
- We should be able to create testbeds with different names, after this fix we can deploy like this-
```
STACK_NAME=prateek-testbed cdk deploy --profile=eks-scalability-dev
```
- Map autoscaling group to Dataplane spec request.
  
### DIFF
--- a/operator/pkg/awsprovider/instances/reconciler.go
+++ b/operator/pkg/awsprovider/instances/reconciler.go
@@ -45,7 +45,7 @@ func NewController(ec2api *awsprovider.EC2, autoscaling *awsprovider.AutoScaling
 }
 
 func (c *Controller) Reconcile(ctx context.Context, dataplane *v1alpha1.DataPlane) error {
-	asg, err := c.getAutoScalingGroup(ctx, AutoScalingGroupNameFor(dataplane.Spec.ClusterName))
+	asg, err := c.getAutoScalingGroup(ctx, AutoScalingGroupNameFor(dataplane))
 	if err != nil {
 		return fmt.Errorf("getting auto scaling group for %v, %w", dataplane.Spec.ClusterName, err)
 	}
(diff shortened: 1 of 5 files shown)


testbed/stack.ts
```diff
@@ -11,8 +11,8 @@ export interface TestbedProps extends cdk.StackProps {
 }
 
 export class Testbed extends cdk.Stack {
-    constructor(scope: cdk.Construct, id: string, props: TestbedProps) {
-        super(scope, id)
+    constructor(scope: cdk.Construct, id: string="testbed", props: TestbedProps) {
```
ellistarn, line 14: Do the defaulting at the outer level.

testbed/bin/testbed.ts
```diff
@@ -2,11 +2,13 @@
...
 import { Testbed } from '../stack'
 
-new Testbed(new cdk.App(), 'testbed', {
+new Testbed(new cdk.App(),
+    process.env.STACK_NAME, {
     env: {
         account: process.env.CDK_DEFAULT_ACCOUNT,
         region: process.env.CDK_DEFAULT_REGION
     },
+    stackName: process.env.STACK_NAME,
```
ellistarn, line 11: Why is this being passed twice? Can't you just use the ID?

operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -243,8 +243,8 @@ func (c *Controller) subnetsForInstances(ctx context.Context, instanceIDs []stri
 	return result, nil
 }
 
-func AutoScalingGroupNameFor(clusterName string) string {
-	return fmt.Sprintf("kit-%s-cluster-dataplane", clusterName)
+func AutoScalingGroupNameFor(dataplane *v1alpha1.DataPlane) string {
+	return fmt.Sprintf("kit-%s-cluster-%s", dataplane.Spec.ClusterName, dataplane.Name)
```
ellistarn, line 247: Consider slashes instead of dashes.

`kit/etarn/nodes`

testbed/bin/testbed.ts
prateekgogia: Using the stack name which a user can pass as an environment variable, is there a way to get the ID for that?

testbed/stack.ts
prateekgogia: I thought defaulting in the constructor is the way to go about it.

hakuna-matatah (approved):

PR #60, opened 2021-11-03
undo passing props as it's not compatible with existing resources that are already created

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
- undo this line from this PR  - https://github.com/awslabs/kubernetes-iteration-toolkit/pull/59/files#diff-0634c4f68fd38278b74a3c1457c186b41b766865cf1e039fbe473988d4faed0eR15

### DIFF
--- a/testbed/stack.ts
+++ b/testbed/stack.ts
@@ -12,7 +12,7 @@ export interface TestbedProps extends cdk.StackProps {
 
 export class Testbed extends cdk.Stack {
     constructor(scope: cdk.Construct, id: string="testbed", props: TestbedProps) {
-        super(scope, id, props)
+        super(scope, id)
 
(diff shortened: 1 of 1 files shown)


ellistarn (approved):

testbed/stack.ts
```diff
@@ -12,7 +12,7 @@ export interface TestbedProps extends cdk.StackProps {
 
 export class Testbed extends cdk.Stack {
     constructor(scope: cdk.Construct, id: string="testbed", props: TestbedProps) {
-        super(scope, id, props)
+        super(scope, id)
```
prateekgogia, line 15: With this change will we be able to deploy multiple testbeds in the same region in one account? IIUC, this includes the CFN stack name but maybe I am missing something here.

PR #62, opened 2021-11-03
IGW routes for KIT CP LB; Custom tags for KIT CP and DP;update karp; remove kitnoderole, instance profile;

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
- Add IGW route for KIT CP APIServer AWS LB access 
- Add custom tags `kit/hostcluster: <cluster-name>-controlplane` tag for Karpenter to scale KIT CP
- Add custom tags `kit/hostcluster: <cluster-name>-dataplane` tag for KIT to discover and launch dataplane nodes
- Update Karp to 0.4.1 to for KIT compatibility
- Remove KITNodeRole and InstanceProfile from KIT addons construct, as we decided to move those resource creation to KIT operator.

### DIFF
--- a/testbed/addons/kit/construct.ts
+++ b/testbed/addons/kit/construct.ts
@@ -52,27 +52,6 @@ export class Kit extends cdk.Construct {
             ],
         }))
 
-        // Node Role
-        const nodeRole = new iam.Role(this, 'kit-node-role', {
-            assumedBy: new iam.ServicePrincipal('ec2.amazonaws.com'),
-            managedPolicies: [
-                iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSWorkerNodePolicy'),
-                iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEC2ContainerRegistryReadOnly'),
-                iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonEKS_CNI_Policy'),
-                iam.ManagedPolicy.fromAwsManagedPolicyName('AmazonSSMManagedInstanceCore')
-            ]
-        })
-
-        props.cluster.awsAuth.addRoleMapping(nodeRole, {
-            username: 'system:node:{{EC2PrivateDNSName}}',
-            groups: ['system:bootstrappers', 'system:nodes']
-        })
-
-        new iam.CfnInstanceProfile(this, 'kit-instance-profile', {
-            roles: [nodeRole.roleName],
-            instanceProfileName: 'KitNodeInstanceProfile'
-        })
-
         // Install kit
         const chart = props.cluster.addHelmChart('kit', {
             chart: 'kit-operator',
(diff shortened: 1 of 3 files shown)


testbed/stack.ts
```diff
@@ -30,6 +30,15 @@ export class Testbed extends cdk.Stack {
                 },
             ],
         });
+
+        //Tag pub subnets for KIT CP
+        const selection = vpc.selectSubnets({
+            subnetType: ec2.SubnetType.PUBLIC
+        });
```
prateekgogia, line 37: Why do we have public subnet for control plane nodes?

testbed/addons/kit/construct.ts
```diff
@@ -92,5 +71,27 @@ export class Kit extends cdk.Construct {
             }
         })
         chart.node.addDependency(ns)
+
+        //Karp Provisioner for kit
```
prateekgogia, line 75: I think this should stay in Karpenter/construct.ts as this is a default provisioner, if we were to create KIT specific provisioner then it would make sense to have it here.

testbed/stack.ts
```diff
@@ -38,26 +47,37 @@ export class Testbed extends cdk.Stack {
                 cidrBlock: `10.${index + 1}.0.0/16`
             });
             let privateSubnet = new ec2.PrivateSubnet(this, `${id}-private-subnet-${index}`, {
-                availabilityZone: cdk.Stack.of(this).availabilityZones[index%cdk.Stack.of(this).availabilityZones.length],
+                availabilityZone: cdk.Stack.of(this).availabilityZones[index % cdk.Stack.of(this).availabilityZones.length],
                 vpcId: vpc.vpcId,
                 cidrBlock: `10.${index + 1}.0.0/16`
             })
             privateSubnet.node.addDependency(additionalCidr);
+            //Tag pub subnets for KIT DP
```
prateekgogia, line 55: nit: pub subnets, these are private subnets.

testbed/stack.ts
```diff
@@ -30,6 +30,15 @@ export class Testbed extends cdk.Stack {
                 },
             ],
         });
+
+        //Tag pub subnets for KIT CP
+        const selection = vpc.selectSubnets({
+            subnetType: ec2.SubnetType.PUBLIC
+        });
+        selection.subnets.forEach(subnet => {
+            Tags.of(subnet).add('kit/hostcluster', `${id}-controlplane`)
```
prateekgogia, line 39: Also will this be a single subnet in one AZ?
I thought about it a little bit more after our discussion and If thats the case, it will be not sufficient because CP nodes are created across different AZs (topology spread constraint)

prateekgogia (changes requested):

testbed/addons/kit/construct.ts
hakuna-matatah: This provisioner is specifically only used for KIT CP, as KIT is leveraging Karp, right ? given it's subnetSelector catered to KIT operator. It makes sense to be here for me, wdyt ?

testbed/stack.ts
hakuna-matatah: It's not mandatory. As per our discussion, KIT has requirement of having non-overlapping subnet for DP and CP, so it's cleaner this way to pick public subnets for CP and private subnets for DP.

testbed/stack.ts
hakuna-matatah: let's discuss offline what the issue is with KIT operator.
 Given megaVPC we can't have just one subnet per AZ as we need lots of IP's. Restricting only one subnet per AZ will limit us from running all megaxl tests that we want.

testbed/stack.ts
prateekgogia: We can also create separate private subnets for CP and DP for a cleaner separation?

prateekgogia:
After discussing offline with Harish, we are currently creating CP in public subnets and across 2 AZs.
We have a follow up issue on this to fix this - https://github.com/awslabs/kubernetes-iteration-toolkit/issues/63

prateekgogia (approved):

PR #65, opened 2021-11-04
Creates IAM roles for DP nodes

prateekgogia:
*Issue #, if available:*

*Description of changes:*
- Creates IAM roles and instance profile for DP nodes
- Fix DNS service IP for guest clusters by providing to instance user-data
- Launch template discovery with filter is replaced with using template name
- Updates the Dataplane CRD in helm charts


operator/docs/kit.cloudformation.yaml
```diff
@@ -1,36 +1,14 @@
...
-      ManagedPolicyArns:
-        - !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonEKSWorkerNodePolicy"
-        - !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonEKS_CNI_Policy"
-        - !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
-        - !Sub "arn:${AWS::Partition}:iam::aws:policy/AmazonSSMManagedInstanceCore"
   KitControllerPolicy:
     Type: "AWS::IAM::ManagedPolicy"
     Properties:
-      ManagedPolicyName: "KitControllerPolicy"
+      ManagedPolicyName: !Sub "KitControllerPolicy-${ClusterName}"
```
hakuna-matatah, line 11: why would you need a `cluster-name` for the policy ?

operator/README.md
```diff
@@ -40,7 +41,7 @@ KIT uses the [operator pattern](https://kubernetes.io/docs/concepts/extend-kuber
     --name kit-controller \
     --namespace kit \
     --cluster ${SUBSTRATE_CLUSTER_NAME} \
-    --attach-policy-arn arn:aws:iam::${AWS_ACCOUNT_ID}:policy/KitControllerPolicy \
+    --attach-policy-arn arn:aws:iam::${AWS_ACCOUNT_ID}:policy/KitControllerPolicy-${SUBSTRATE_CLUSTER_NAME} \
```
hakuna-matatah, line 44: why `cluster-name` ?

operator/docs/kit.cloudformation.yaml
```diff
@@ -49,10 +27,21 @@ Resources:
               - "autoscaling:DeleteAutoScalingGroup"
               - "autoscaling:UpdateAutoScalingGroup"
               - "autoscaling:SetDesiredCapacity"
+              - "iam:CreateRole"
```
hakuna-matatah, line 30: If you need these to create role, then could you also these additional permissions here to maintain parity- https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/testbed/addons/kit/construct.ts#L33-L51

operator/charts/kit-operator/templates/data-plane-crd.yaml
```diff
@@ -35,10 +35,35 @@ spec:
             type: object
           spec:
             properties:
+              allocationStrategy:
```
hakuna-matatah, line 38: These additions to CRD forces us to change in KIT tekton specs here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tasks/setup/kit-cluster/default.yaml#L38-L95

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
```diff
@@ -82,3 +84,8 @@ type Instances struct {
 func (c *ControlPlane) ClusterName() string {
 	return c.Name
 }
+
+type ReconcileFinalize interface {
```
hakuna-matatah, line 88: wondering if we could come up with a better name instead of `ReconcileFinalize` ? Thoughts ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"knative.dev/pkg/ptr"
+)
+
+type Controller struct {
+	iam        *awsprovider.IAM
+	kubeClient *kubeprovider.Client
+}
+
+// NewController returns a controller for managing LaunchTemplates and ASG in AWS
```
hakuna-matatah, line 37: nit: I think its a copy paste error :D

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+func NewController(iam *awsprovider.IAM, client *kubeprovider.Client) *Controller {
+	return &Controller{iam: iam, kubeClient: client}
+}
+
+var (
+	desiredPolicies = []string{
```
hakuna-matatah, line 43: variable name sounds too generic here. In reality these are policies required for `kitnoderole`

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("removing role from instance profile, %w", err)
+	}
+	_, err = c.iam.DeleteInstanceProfileWithContext(ctx, &iam.DeleteInstanceProfileInput{
+		InstanceProfileName: aws.String(KitNodeInstanceProfileNameFor(controlPlane.ClusterName())),
+	})
+	if err != nil && !errors.IsIAMObjectDoNotExist(err) {
+		return fmt.Errorf("deleting instance profile, %w", err)
+	}
+
+	for _, policy := range desiredPolicies {
```
hakuna-matatah, line 90: why would you want to detach policies before deleting the role ? why can't you just delete the role straight away ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		if err := role.attachPolicy(ctx, policy, controlPlane.ClusterName()); err != nil {
+			return fmt.Errorf("attaching policies to role, %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (c *Controller) Finalize(ctx context.Context, controlPlane *apis.ControlPlane) error {
+	_, err := c.iam.RemoveRoleFromInstanceProfileWithContext(ctx, &iam.RemoveRoleFromInstanceProfileInput{
```
hakuna-matatah, line 76: is this operation really required ? can't you just delete instance profile and role ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("deleting role, %w", err)
+	}
+	zap.S().Infof("[%s] Deleted IAM Role %v and instance profile",
+		controlPlane.ClusterName(), KitNodeRoleNameFor(controlPlane.ClusterName()))
+	return nil
+}
+
+func (c *Controller) getRole(ctx context.Context, controlPlane *apis.ControlPlane) (*role, error) {
+	roleOutput, err := c.iam.GetRoleWithContext(ctx, &iam.GetRoleInput{
+		RoleName: aws.String(KitNodeRoleNameFor(controlPlane.ClusterName())),
```
hakuna-matatah, line 111: function name sounds generic but this function is only good for `kitNodeROle` based on this line. Instead of passing `controlPlane *apis.ControlPlane` as func param, i would recommend passing `rolename` as arg to this function. wdyt ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (c *Controller) Finalize(ctx context.Context, controlPlane *apis.ControlPlane) error {
+	_, err := c.iam.RemoveRoleFromInstanceProfileWithContext(ctx, &iam.RemoveRoleFromInstanceProfileInput{
+		InstanceProfileName: aws.String(KitNodeInstanceProfileNameFor(controlPlane.ClusterName())),
+		RoleName:            aws.String(KitNodeRoleNameFor(controlPlane.ClusterName())),
```
hakuna-matatah, line 78: function name sounds generic but this function is only good for `kitNodeROle` based on this line. Instead of passing `controlPlane *apis.ControlPlane` as func param, i would recommend passing `rolename` as arg to this function. Thoughts ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return &role{iam: c.iam, Role: roleOutput.Role}, nil
+}
+
+func (c *Controller) createRole(ctx context.Context, controlPlane *apis.ControlPlane) (*role, error) {
+	roleOutput, err := c.iam.CreateRole(&iam.CreateRoleInput{
+		AssumeRolePolicyDocument: aws.String(assumeRolePolicyDocument),
+		Description:              aws.String("Role assumed by dataplane nodes created by KIT operated"),
+		RoleName:                 aws.String(KitNodeRoleNameFor(controlPlane.ClusterName())),
+		Tags:                     generateRoleTags(controlPlane.ClusterName()),
```
hakuna-matatah, line 124: function name sounds generic but this function is only good for `kitNodeROle` based on this line. Instead of passing `controlPlane *apis.ControlPlane` as func param, i would recommend passing `rolename` as arg to this function. Thoughts ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+	return &role{iam: c.iam, Role: roleOutput.Role}, err
+}
+
+type role struct {
+	iam *awsprovider.IAM
+	*iam.Role
+}
+
+func (r *role) addRoleToInstanceProfile(ctx context.Context, controlPlane *apis.ControlPlane) error {
+	profile, err := createInstanceProfile(ctx, r.iam, KitNodeInstanceProfileNameFor(controlPlane.ClusterName()))
```
hakuna-matatah, line 135: function name sounds generic but this function is only good for `kitNodeROle` based on this line. Instead of passing `controlPlane *apis.ControlPlane` as func param, i would recommend passing `rolename` as arg to this function. Thoughts ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+		InstanceProfileName: aws.String(KitNodeInstanceProfileNameFor(controlPlane.ClusterName())),
+		RoleName:            aws.String(KitNodeRoleNameFor(controlPlane.ClusterName())),
+	})
+	return err
+}
+
+func (r *role) attachPolicy(ctx context.Context, policyARN, name string) error {
+	_, err := r.iam.AttachRolePolicy(&iam.AttachRolePolicyInput{
+		PolicyArn: aws.String(policyARN),
+		RoleName:  aws.String(KitNodeRoleNameFor(name)),
```
hakuna-matatah, line 157: function name sounds generic but this function is only good for `kitNodeROle` based on this line.

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+const assumeRolePolicyDocument = `{
+	"Version": "2012-10-17",
+	"Statement": [
+		{
+			"Action": "sts:AssumeRole",
+			"Effect": "Allow",
+			"Principal": {
+				"Service": "ec2.amazonaws.com"
```
hakuna-matatah, line 206: could you add details on why you need this ?

operator/pkg/awsprovider/launchtemplate/reconciler.go
```diff
@@ -212,6 +208,7 @@ var (
 #!/bin/bash
 yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm
 /etc/eks/bootstrap.sh %s.%s \
+    --dns-cluster-ip %s \
```
hakuna-matatah, line 211: could you add the issue here in comments as todo ?

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
```diff
@@ -82,3 +84,8 @@ type Instances struct {
...
+	Reconcile(context.Context, *ControlPlane) error
+	Finalize(context.Context, *ControlPlane) error
```
hakuna-matatah, line 90: Looks like finalize doesn't make sense in many/most of your controllers like etcd and addons (kubeproxy and coredns) ? Wondering why you wanted it to be here ?

operator/pkg/controllers/controlplane/controlplane.go
```diff
@@ -38,10 +37,10 @@ type controlPlane struct {
 }
 
 // NewController returns a controller for managing controlPlane components of the cluster
-func NewController(kubeClient client.Client, account awsprovider.AccountMetadata) *controlPlane {
+func NewController(kubeClient client.Client, account awsprovider.AccountMetadata, iamProvider v1alpha1.ReconcileFinalize) *controlPlane {
 	return &controlPlane{
 		etcdController:   etcd.New(kubeprovider.New(kubeClient)),
-		masterController: master.New(kubeprovider.New(kubeClient), account),
+		masterController: master.New(kubeprovider.New(kubeClient), account, iamProvider),
 		addonsController: addons.New(kubeprovider.New(kubeClient)),
```
hakuna-matatah, line 44: Also, on a different note, this doesn't make sense to me to be part of controlplane. Any reason why you have added this to controlplane, instead of dataplane ?

operator/pkg/controllers/addons/addons.go
```diff
@@ -50,7 +49,7 @@ func (c *Controller) Reconcile(ctx context.Context, controlPlane *v1alpha1.Contr
 		return err
 	}
 	// reconcile addons to the guest cluster
-	for _, resource := range []reconciler.Interface{
+	for _, resource := range []v1alpha1.ReconcileFinalize{
```
hakuna-matatah, line 52: why do you have to change to this interface defined in controlplane api for addons ?

operator/pkg/controllers/controlplane/controlplane.go
```diff
@@ -58,7 +57,7 @@ func (c *controlPlane) For() controllers.Object {
 
 // Reconcile will reconcile all the components running on the control plane
 func (c *controlPlane) Reconcile(ctx context.Context, object controllers.Object) (res *reconcile.Result, err error) {
-	for _, resource := range []reconciler.Interface{
+	for _, resource := range []v1alpha1.ReconcileFinalize{
```
hakuna-matatah, line 60: addons and etcd doesn't really need this interface IIUC.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: No it fails to delete without doing this.

operator/README.md
prateekgogia: same as above.

operator/charts/kit-operator/templates/data-plane-crd.yaml
prateekgogia: We will have to update the spec anyways to include the subnet selector and most of this changes are adding description except the subnetSelector being a new field.

operator/docs/kit.cloudformation.yaml
prateekgogia: I was thinking of cutting a separate PR, but I can push these changes to this one.

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
prateekgogia: It's more like a Lifecycle of the object, so I kept it simple to include the operations interface does. Open to suggestions.

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
prateekgogia: Because I need to pass `ReconcileFinalize` interface for tests else there is no way to provide a fake provider. For some of the controller its a noop.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: not clear which function you are refering to here `KitNodeRoleNameFor`?

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: No it fails to delete the role directly.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: I think this is the same as above.

operator/pkg/awsprovider/launchtemplate/reconciler.go
prateekgogia: I have added it where the const is defined

operator/pkg/controllers/addons/addons.go
prateekgogia: I need and interface which has both Reconcile and Finalize for the tests to succeed.

operator/pkg/controllers/controlplane/controlplane.go
prateekgogia: Addons you mean?

operator/pkg/controllers/controlplane/controlplane.go
prateekgogia: They all have the same logic for Reconcile and Finalize

operator/README.md
prateekgogia: I will fix this one.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: Will update the function param.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: Will update the function param.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: update params to provide roleName instead of clusterName

operator/docs/kit.cloudformation.yaml
hakuna-matatah: >this is the issue we were facing when there is an existing testbed in an account and you can't deploy another because CDK can't skip existing roles/policies.

We discussed offline, this isn't the issue we were facing earlier. It was for KitNodeInstanceRole as KitNodeInstanceProfile is unique.

operator/charts/kit-operator/templates/data-plane-crd.yaml
hakuna-matatah: do you have an example spec added in example folder leveraging all these new fields to refer for Tekton tasks  and also will help other customers ?

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,211 @@
...
+	return &role{iam: c.iam, Role: roleOutput.Role}, err
+}
+
+type role struct {
+	iam *awsprovider.IAM
+	*iam.Role
+}
+
+func (r *role) addRoleToInstanceProfile(ctx context.Context, clusterName string) error {
+	profile, err := createInstanceProfile(ctx, r.iam, KitNodeInstanceProfileNameFor(clusterName))
```
hakuna-matatah, line 135: could you take `InstanceProfileName` as param to make it compliant with your function name as it's generic.

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,211 @@
...
+		return fmt.Errorf("deleting role, %w", err)
+	}
+	zap.S().Infof("[%s] Deleted IAM Role %v and instance profile",
+		controlPlane.ClusterName(), KitNodeRoleNameFor(controlPlane.ClusterName()))
+	return nil
+}
+
+func (c *Controller) getRole(ctx context.Context, clusterName string) (*role, error) {
+	roleOutput, err := c.iam.GetRoleWithContext(ctx, &iam.GetRoleInput{
+		RoleName: aws.String(KitNodeRoleNameFor(clusterName)),
```
hakuna-matatah, line 111: i think you are still not taking rolename. This function is still tied to kitnoderole though name sounds generic.

operator/pkg/awsprovider/iam/reconciler.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, fmt.Errorf("getting iam role %v, %w", KitNodeRoleNameFor(clusterName), err)
+	}
+	return &role{iam: c.iam, Role: roleOutput.Role}, nil
+}
+
+func (c *Controller) createRole(ctx context.Context, clusterName string) (*role, error) {
+	roleOutput, err := c.iam.CreateRole(&iam.CreateRoleInput{
+		AssumeRolePolicyDocument: aws.String(assumeRolePolicyDocument),
+		Description:              aws.String("Role assumed by dataplane nodes created by KIT operated"),
+		RoleName:                 aws.String(KitNodeRoleNameFor(clusterName)),
```
hakuna-matatah, line 123: I think here as well, you might need to accept rolename to make it compliant with your generic func name.

testbed/addons/kit/construct.ts
```diff
@@ -41,13 +41,24 @@ export class Kit extends cdk.Construct {
                         "autoscaling:DeleteAutoScalingGroup",
                         "autoscaling:UpdateAutoScalingGroup",
                         "autoscaling:SetDesiredCapacity",
+                        "iam:CreateRole",
```
hakuna-matatah, line 44: Nice, thank you sir.

operator/pkg/awsprovider/iam/reconciler.go
prateekgogia: Here I also need clusterName to generate tags. So its hard to just accept the role name in params

operator/charts/kit-operator/templates/data-plane-crd.yaml
prateekgogia: Yesh its there - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/operator/docs/examples/dataplane.yaml#L8

hakuna-matatah (approved):
lgtm

PR #68, opened 2021-11-05
increase default cidr to accommodate all default host controller pods.

hakuna-matatah:
*Issue #, if available:*

*Description of changes:*
- /28 is not enough to accommodate pods like KIT. This will become a problem as we add more addons to the testbed anyway, so moved to /24.

### DIFF
--- a/testbed/stack.ts
+++ b/testbed/stack.ts
@@ -21,12 +21,12 @@ export class Testbed extends cdk.Stack {
                 {
                     name: 'pub-subnet-1',
                     subnetType: ec2.SubnetType.PUBLIC,
-                    cidrMask: 28,
+                    cidrMask: 24,
                 },
                 {
                     name: 'priv-subnet-1',
                     subnetType: ec2.SubnetType.PRIVATE_WITH_NAT,
-                    cidrMask: 28,
+                    cidrMask: 24,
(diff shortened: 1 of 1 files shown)


testbed/stack.ts
```diff
@@ -21,12 +21,12 @@ export class Testbed extends cdk.Stack {
...
                     name: 'pub-subnet-1',
                     subnetType: ec2.SubnetType.PUBLIC,
-                    cidrMask: 28,
+                    cidrMask: 24,
                 },
                 {
                     name: 'priv-subnet-1',
                     subnetType: ec2.SubnetType.PRIVATE_WITH_NAT,
-                    cidrMask: 28,
+                    cidrMask: 24,
```
prateekgogia, line 29: We should also check that the node group created to have private subnets, currently they get 4 subnets assigned. I am not sure if they get public and private both.

prateekgogia (approved):

testbed/stack.ts
hakuna-matatah: It selects only private subnets from VPC for default nodegroups, see here https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/testbed/stack.ts#L69-L70

Issue #72, opened 2021-11-18
[Operator] Dataplane spec is mandating to have allocationStrategy in the spec


hakuna-matatah:
- Dataplane spec is mandating to have allocationStrategy in the spec. If not provided it will run into the following issue:
 ```
message: "creating auto scaling group for s-9a7d13d2-e635-4418-8852-7d528e66fa69,
      ValidationError: 1 validation error detected: Value '' at 'mixedInstancesPolicy.launchStrategy.onDemandAllocationStrategy'
      failed to satisfy constraint: Member must have length greater than or equal
      to 1\n\tstatus code: 400, request id: 270dd3de-8aa9-4ff2-baeb-81ac9cfd1fd0"
    status: "False"
    type: Active
```

We need to look into why defaulting doesn't work. If we plan to remove defaulting, we need to update the example spec [here](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/operator/docs/examples/dataplane.yaml)] with allocation strategy

prateekgogia:
Fixed in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/75

PR #75, opened 2021-11-22
add dataplane webhooks and fix dp finalizer

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Dataplane mutating and validating webhooks were not enabled, this PR enables them.
 - Fixed Dataplane finalizer where if an ASG had not been created due to a resource error in the DP (the specific case I came across was when the `allocationStrategy` was not set and not defaulted since the webhooks weren't wired up), the finalizer would fail since the ASG deletion would error. It now checks if the ASG exists before trying to delete it.
 - `make apply` and `make delete` use the templated helm chart now. And CRDs are in the proper directory now. 


prateekgogia (approved):
Nicely done! LGTM!
Fixes https://github.com/awslabs/kubernetes-iteration-toolkit/issues/72

PR #83, opened 2021-12-07
Support custom host-clustername for KIT tekton tasks; Tear down KIT clusters regardless of success/fail tests; Upgrade KIT to `0.0.4`

hakuna-matatah:
- Update KIT to `0.0.4`
- Support custom host-clustername for KIT tekton tasks
- Tear down KIT clusters regardless of success/fail tests
- Adhere tekton to KIT requirements like wait for NLB creation, subnet-selectors etc (long standing changes on fork need to be merged

### DIFF
--- a/testbed/stack.ts
+++ b/testbed/stack.ts
@@ -63,7 +63,6 @@ export class Testbed extends cdk.Stack {
             version: eks.KubernetesVersion.V1_19,
             defaultCapacity: 0,
         })
-
         cluster.addNodegroupCapacity('node-group', {
             nodegroupName: 'default',
             subnets: vpc.selectSubnets({
@@ -80,9 +79,12 @@ export class Testbed extends cdk.Stack {
             }),
         })
         // service account used by tekton workflows.
-        cluster.addServiceAccount('test-executor', { name: 'test-executor' })
-            .role.addManagedPolicy({ managedPolicyArn: 'arn:aws:iam::aws:policy/AdministratorAccess' })
-
+        let sa = cluster.addServiceAccount('test-executor', { name: 'test-executor' })
+        sa.role.addManagedPolicy({ managedPolicyArn: 'arn:aws:iam::aws:policy/AdministratorAccess' })
+        cluster.awsAuth.addRoleMapping(sa.role, {
+            username: sa.role.roleName,
+            groups: ['system:masters']
+        })
         new Addons(this, `${id}-addons`, { cluster: cluster, repositories: props.repositories })
 
         // Tag all resources for discovery by Karpenter
--- a/tests/pipelines/kit/upstream-load.yaml
+++ b/tests/pipelines/kit/upstream-load.yaml
@@ -9,6 +9,9 @@ spec:
     - name: source
     - name: results
   params:
+  - name: host-cluster-name
+    description: The name of the Host cluster on which you spin up KIT Guest cluster.
+    default: "testbed"
   - name: cluster-name
     description: The name of the kit cluster you want to spin.
   - name: host-cluster-region
@@ -35,9 +38,9 @@ spec:
   - name: scheduler-qps
     default: "20"
     description: Kubernetes-Scheduler QPS setting
-  - name: node-count
+  - name: node_count
     default: "1000"
-    description: desired node count for Dataplane.
+    description: desired node count for Dataplane, min is 1000 to create DP nodes.
   - name: giturl
     description: "git url to clone the package"
     default: https://github.com/kubernetes/perf-tests.git
@@ -57,6 +60,8 @@ spec:
     taskRef:
       name: kit-cluster-create
     params:
+      - name: host-cluster-name
+        value: '$(params.host-cluster-name)'   
       - name: cluster-name
         value: '$(params.cluster-name)'
       - name: host-cluster-region
@@ -75,8 +80,8 @@ spec:
         value: '$(params.kcm-qps)' 
       - name: scheduler-qps
         value: '$(params.scheduler-qps)'
-      - name: node-count
-        value: '$(params.node-count)'      
+      - name: node_count
+        value: '$(params.node_count)'      
     workspaces:
       - name: config    
         workspace: config
@@ -96,7 +101,7 @@ spec:
       - name: results-bucket
         value: '$(params.results-bucket)'
       - name: nodes
-        value: '$(params.node-count)'
+        value: '$(params.node_count)'
     workspaces:
       - name: source
         workspace: source
@@ -104,10 +109,12 @@ spec:
         workspace: config
       - name: results
         workspace: results
+  finally:      
   - name: teardown
-    runAfter: [generate]
     taskRef:
       name: kit-cluster-teardown
     params:
+      - name: host-cluster-name
+        value: '$(params.host-cluster-name)'        
       - name: cluster-name
         value: '$(params.cluster-name)'
--- a/tests/tasks/generators/clusterloader/load.yaml
+++ b/tests/tasks/generators/clusterloader/load.yaml
@@ -42,13 +42,16 @@ spec:
       MEDIUM_STATEFUL_SETS_PER_NAMESPACE: 0
       # we are not testing PVS at this point
       CL2_ENABLE_PVS: false
+      PROMETHEUS_SCRAPE_APISERVER_ONLY: true
       PROMETHEUS_SCRAPE_KUBE_PROXY: false
       ENABLE_SYSTEM_POD_METRICS: false
       NODE_MODE: master 
       EOL
+      cat $(workspaces.source.path)/overrides.yaml
+      cp $(workspaces.source.path)/overrides.yaml $(workspaces.results.path)/overrides.yaml
   - name: validate-cluster
     image: amazon/aws-cli
-    workingDir: $(workspaces.source.path)
+    workingDir: $(workspaces.config.path)
     script: |
       mkdir -p /root/.kube/
       cp $(workspaces.config.path)/kubeconfig /root/.kube/config
@@ -63,19 +66,19 @@ spec:
       # end
   - name: run-loadtest
     image: 197575167141.dkr.ecr.us-west-2.amazonaws.com/clusterloader2:76e3fd7
-    command: ["./clusterloader"]
-    args: 
-    - --kubeconfig=$(workspaces.config.path)/kubeconfig
-    - --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml
-    - --testoverrides=$(workspaces.source.path)/overrides.yaml
-    - --nodes=$(params.nodes) 
-    - --provider=eks 
-    - --report-dir=$(workspaces.results.path)
-    - --alsologtostderr
-    env:
-    - name: ENABLE_EXEC_SERVICE
-      value: "false"
+    script: |
+      ENABLE_EXEC_SERVICE=false ./clusterloader --kubeconfig=$(workspaces.config.path)/kubeconfig --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr
+      #To run the next step regardless of success/failure of the test
+      TEST_EXIT_CODE=$?
+      if [ $TEST_EXIT_CODE != 0 ]; then
+        exit 0
+      fi
+    timeout: 30000s
   - name: upload-results
     image: amazon/aws-cli
     workingDir: $(workspaces.results.path)
-    command: ["aws s3 cp . s3://$(params.results-bucket)/  --recursive "]
+    script: |
+      aws sts get-caller-identity
+      # we expect to see all files from loadtest that clusterloader2 outputs here in this dir
+      ls -larth
+      aws s3 cp . s3://$(params.results-bucket)/  --recursive
--- a/tests/tasks/setup/kit-cluster/default.yaml
+++ b/tests/tasks/setup/kit-cluster/default.yaml
@@ -9,6 +9,9 @@ spec:
     This Task can be used to create an KIT cluster in an AWS account and write a kubeconfig to a desired location that
     can be used by other tasks (in a context with kubectl) to make requests to the cluster.
   params:
+  - name: host-cluster-name
+    description: The name of the Host cluster on which you spin up KIT Guest cluster.
+    default: "testbed"
   - name: cluster-name
     description: The name of the KIT cluster you want to spin.
   - name: host-cluster-region
@@ -35,7 +38,7 @@ spec:
   - name: scheduler-qps
     default: "20"
     description: Kubernetes-Scheduler QPS setting
-  - name: node-count
+  - name: node_count
     default: "1000"
     description: desired node count for Dataplane.
   workspaces:
@@ -49,7 +52,7 @@ spec:
       curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
       install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
       kubectl version
-      aws eks update-kubeconfig --name testbed --region $(params.host-cluster-region)
+      aws eks update-kubeconfig --name $(params.host-cluster-name) --region $(params.host-cluster-region)
       kubectl config current-context
       #Provision a control plane for the guest cluster
       cat > "$(workspaces.config.path)/cp.yaml" <<EOL
@@ -58,7 +61,7 @@ spec:
       metadata:
         name: $(params.cluster-name) # Desired Cluster Name
       spec:
-        kubernetesVersion: $(params.version)
+        kubernetesVersion: "$(params.version)"
         master:
           apiServer:
             spec:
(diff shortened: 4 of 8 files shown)


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -42,13 +42,15 @@ spec:
       MEDIUM_STATEFUL_SETS_PER_NAMESPACE: 0
       # we are not testing PVS at this point
       CL2_ENABLE_PVS: false
+      PROMETHEUS_SCRAPE_APISERVER_ONLY: true
       PROMETHEUS_SCRAPE_KUBE_PROXY: false
       ENABLE_SYSTEM_POD_METRICS: false
       NODE_MODE: master 
       EOL
+      cat $(workspaces.source.path)/overrides.yaml
```
prateekgogia, line 50: Do we need this for debugging?
Can we include this info in the results uplaoded to S3?

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -63,19 +65,18 @@ spec:
...
-    - name: ENABLE_EXEC_SERVICE
-      value: "false"
+    script: |
+      ENABLE_EXEC_SERVICE=false ./clusterloader --kubeconfig=$(workspaces.config.path)/kubeconfig --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr
+      #To run the next step regardless of success/failure of the test
+      TEST_EXIT_CODE=$?
+      if [ $TEST_EXIT_CODE != 0 ]; then
+        exit 0
+      fi
+    timeout: 30000s
```
prateekgogia, line 75: This also is a very long time to run these tests

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -63,19 +65,18 @@ spec:
...
-    - --testoverrides=$(workspaces.source.path)/overrides.yaml
-    - --nodes=$(params.nodes) 
-    - --provider=eks 
-    - --report-dir=$(workspaces.results.path)
-    - --alsologtostderr
-    env:
-    - name: ENABLE_EXEC_SERVICE
-      value: "false"
+    script: |
+      ENABLE_EXEC_SERVICE=false ./clusterloader --kubeconfig=$(workspaces.config.path)/kubeconfig --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr
```
prateekgogia, line 69: Do we need to pass `--provider=eks` when testing on KIT as well?

tests/triggers/rbac.yaml
```diff
@@ -64,3 +64,21 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: tekton-triggers
+---
+apiVersion: v1
```
prateekgogia, line 68: I am guessing these are created in guest cluster?

tests/triggers/kit/small/upstream-load.yaml
```diff
@@ -44,18 +44,25 @@ spec:
     metadata:
       generateName: kit-small-upstream-load-
     spec:
+      timeout: "9h0m0s"
       pipelineRef:
-        name: loadtest
+        name: kitloadtest
```
prateekgogia, line 49: rename to `kitSmallUpstreamLoadTest`?

tests/triggers/eks/small/upstream-load.yaml
```diff
@@ -4,7 +4,7 @@ kind: CronJob
 metadata:
   name: eks-small-upstream-load
 spec:
-  schedule: "0 8 * * *"
+  schedule: "0 0 1 * *"
```
prateekgogia, line 7: Are we planning to run this at every month's frequency?

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -63,19 +65,18 @@ spec:
...
+        exit 0
+      fi
+    timeout: 30000s
   - name: upload-results
     image: amazon/aws-cli
     workingDir: $(workspaces.results.path)
-    command: ["aws s3 cp . s3://$(params.results-bucket)/  --recursive "]
+    script: |
+      aws sts get-caller-identity
+      ls -larth
```
prateekgogia, line 81: Can we add a comment here, as to what all we expect to get uploaded to S3?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
               - name: scheduler
                 args:
                   - --kube-api-qps=$(params.scheduler-qps)
+        etcd:
+          spec:
+            nodeSelector: 
+              node.kubernetes.io/instance-type: $(params.cp-instance-type)
```
prateekgogia, line 90: Should this be `etcd-instance-type`? We will need to have a new param for this.

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+            containers: 
+            - name: etcd
+              args:
+                - --quota-backend-bytes=8589934592
```
prateekgogia, line 94: KIT-operator already has this configuration, so we don't need to pass this flag

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+      #validate 
+      kubectl get pods -A
       kubectl create -f $(workspaces.config.path)/cp.yaml
-
-      cat > "$(workspaces.config.path)/dp.yaml" <<EOL
+      nodes=$(params.node_count)
+      #TODO: Assumption is always nodecount is greater than or equal to 1000 for scalability testing purposes.
+      asgs=$((nodes/1000))
+      asg_name=$(params.cluster-name)-nodes
+      for i in $(seq 1 $asgs)
```
prateekgogia, line 108: Can we fix the indentation for this for loop, its harder to read the way it is right now?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+      do
+      filename="$(workspaces.config.path)/dp$i"."yaml"
+      echo $filename
```
prateekgogia, line 111: Doesn't seem like `filename` is being used anywhere else. Can we remove this variable?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+      # kubectl config set-credentials tekton --token=$TOKEN
+      # kubectl config set-context --current --user=tekton
+      #validate 
+      kubectl get pods -A
       kubectl create -f $(workspaces.config.path)/cp.yaml
-
-      cat > "$(workspaces.config.path)/dp.yaml" <<EOL
+      nodes=$(params.node_count)
+      #TODO: Assumption is always nodecount is greater than or equal to 1000 for scalability testing purposes.
+      asgs=$((nodes/1000))
```
prateekgogia, line 106: What happens when nodes=1100, asgs will be 1.
The for loop will run only once with i=1? Where as we need to create atleast 2 ASGs in this case.

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
-        nodeCount: $(params.node-count)
+        nodeCount: 1000 #Max nodes we can accomodate per ASG in KIT today
+        subnetSelector:
+          kit/hostcluster: "$(params.host-cluster-name)-dataplane"
+        allocationStrategy: "flexible"
+        instanceTypes: #defaulting to nitro-instances for now
+          - m5.2xlarge
+          - m5.4xlarge
+          - m4.2xlarge
+          - m4.4xlarge
```
prateekgogia, line 127: Why not m5.large or xlarge, 2xlarge and 4xlarge for scale testing seems to big when creating 10K+ nodes.
Also why are we defaulting to nitro instances, is there any benefit?

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
-
+      ls -larth $(workspaces.config.path)
+      kubectl create -f "$(workspaces.config.path)/dp$i"."yaml"
+      done
+      # sleep for kude-admin config secret for kit guest cluster to appear
+      sleep 30
       #Get the admin KUBECONFIG for the guest cluster from the substrate cluster
       kubectl get secret $(params.cluster-name)-kube-admin-config -ojsonpath='{.data.config}' | base64 -d > $(workspaces.config.path)/kubeconfig
+      # wait for NLB APIServer endpoint to be available
+      sleep 500
```
prateekgogia, line 138: Lets keep it at 300 seconds

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
       #Deploy CNI plugin to the guest cluster for the nodes to be ready.
       #Todo: CNI installation only works for `us-west-2`, yet to decide how to tackle other regions.
       kubectl --kubeconfig=$(workspaces.config.path)/kubeconfig apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/master/config/v1.9/aws-k8s-cni.yaml
-      sleep 300
+      sleep 60
+      kubectl --kubeconfig=$(workspaces.config.path)/kubeconfig get pods -A 
+      # for nodes to become ready 
+      sleep $((nodes/6))
```
prateekgogia, line 145: Not sure why are we sleeping here as this task is about bringing up cluster and we don't check the DP nodes state after that and what the reasoning behind /6?

prateekgogia (changes requested):

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+      # TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+      # kubectl config set-credentials tekton --token=$TOKEN
+      # kubectl config set-context --current --user=tekton
+      #validate 
+      kubectl get pods -A
       kubectl create -f $(workspaces.config.path)/cp.yaml
-
-      cat > "$(workspaces.config.path)/dp.yaml" <<EOL
+      nodes=$(params.node_count)
+      #TODO: Assumption is always nodecount is greater than or equal to 1000 for scalability testing purposes.
```
haugenj, line 105: Do we need to make this assumption? My impression/hope was that this task would be reusable for other pipelines, but it seems like you're locking it into this scale testing case

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,62 @@ spec:
...
+            containers: 
+            - name: etcd
+              args:
+                - --quota-backend-bytes=8589934592
       EOL
+      aws sts get-caller-identity
+      #Use service account attached to pod
+      # TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
+      # kubectl config set-credentials tekton --token=$TOKEN
+      # kubectl config set-context --current --user=tekton
```
haugenj, line 100: are these intentionally commented out?

tests/tasks/setup/kit-cluster/default.yaml
haugenj: can the data plane instance types be parameterized?

tests/pipelines/kit/upstream-load.yaml
```diff
@@ -35,9 +38,9 @@ spec:
   - name: scheduler-qps
     default: "20"
     description: Kubernetes-Scheduler QPS setting
-  - name: node-count
+  - name: node_count
```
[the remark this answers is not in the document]
hakuna-matatah: it was causing  issues with arithmetic operations using this var.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -42,13 +42,15 @@ spec:
       MEDIUM_STATEFUL_SETS_PER_NAMESPACE: 0
       # we are not testing PVS at this point
       CL2_ENABLE_PVS: false
+      PROMETHEUS_SCRAPE_APISERVER_ONLY: true
```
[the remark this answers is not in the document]
hakuna-matatah: test sets up its own prometheus stack.

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: yes, we need those for debugging. Good point, we need to upload that to s3. It was a to-do on my end.

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: Trust me, it ran out even on this timeout because of instability of KIT controlplane when testing megaxl load :D

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: Yes, `provider` is mandatory IIRC. Also, if we don't provide `eks` and provide something else, it will turn on other default features that works for other providers that we mention and we don't want to get into that business at this point, given our setup :D ( `eks` is safe bet because it asks for less feature-set in configuration)

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: Currently we have same param for both. Atleast in my testing, I only use same instance type for CPIs and etcds. But i see why we might need it for other testing purposes. I will add that in next PR.

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: You mean that's the default value in `KIT operator` ? Or is it only hardcoded to just that value ?

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: good catch. that's was from my debugging. I will strip them out.

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: are you saying we want to support lesser than 1000 nodes for any testing that we do with KIT ?

tests/triggers/eks/small/upstream-load.yaml
hakuna-matatah: Yes, for now. Which is why I moved it to monthly because we want to do only testing with KIT for now.

tests/triggers/eks/small/upstream-load.yaml
```diff
@@ -58,4 +58,4 @@ spec:
       - name: cluster-name
         value: "eks-small-upstream-load-$(context.pipelineRun.uid)"
       - name: results-bucket
-        value: "eks-scalability-dev/$(date +'%Y/%m/%d/%HH/%mm)/$(context.pipelineRun.uid)"
+        value: "eks-scalability-dev/$(context.pipelineRun.uid)"
```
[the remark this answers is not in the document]
hakuna-matatah: each upload will have the date and time when it's uploaded to s3 already  :D

tests/triggers/kit/small/upstream-load.yaml
```diff
@@ -44,18 +44,25 @@ spec:
     metadata:
       generateName: kit-small-upstream-load-
     spec:
+      timeout: "9h0m0s"
```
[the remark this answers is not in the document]
hakuna-matatah: yes. I have seen cases where tests ran this long, it's a safe bet for now :D

tests/triggers/kit/small/upstream-load.yaml
hakuna-matatah: Pipeline is going to be same for `small` , `medium` , `large` `XL` etc. We will be reusing it, that's one the main purposes of leveraging tekton pipelines  :D

tests/triggers/rbac.yaml
hakuna-matatah: No, on host-cluster.

testbed/stack.ts
```diff
@@ -80,9 +79,12 @@ export class Testbed extends cdk.Stack {
...
         })
         // service account used by tekton workflows.
-        cluster.addServiceAccount('test-executor', { name: 'test-executor' })
-            .role.addManagedPolicy({ managedPolicyArn: 'arn:aws:iam::aws:policy/AdministratorAccess' })
-
+        let sa = cluster.addServiceAccount('test-executor', { name: 'test-executor' })
+        sa.role.addManagedPolicy({ managedPolicyArn: 'arn:aws:iam::aws:policy/AdministratorAccess' })
+        cluster.awsAuth.addRoleMapping(sa.role, {
+            username: sa.role.roleName,
+            groups: ['system:bootstrappers', 'system:masters']
```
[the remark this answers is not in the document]
hakuna-matatah: Copy paste error from other roles i had :D.

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: I was thinking of parameterizing it in future PR's. Let's take that out of scope for this PR. 
I defaulted to min `2x` large because we can go up to `pod-denisty` of 50 or so if we want to in our testing :D 
I will tell you offline why I'm going for  specifically `nitro-instances` 😉

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: as part of bringing up the cluster we also spin up dataplane nodes/asgs. So, we before we hand it off to further tasks(if any) to do any other operations on this cluster, this will make sure it takes enough time for the nodes to join cluster, so for  every `1000` nodes it's like 2.7 mins roughly. Based on my experience this is safest and reasonable number i could come up with.

tests/tasks/setup/kit-cluster/default.yaml
haugenj: Yeah, I don't see why KIT testing needs to be restricted to clusters of 1000+ nodes, but maybe I'm missing some context

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: Below line already adds that. I think your requirement comment is obsolete dude ^^^

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: A successful test with 10 load throughput would take around 6-7hrs. We need logs and data from the failed tests which could run up to this time. If i see this is causing issues, i'll reduce it in future, I think this is good for now.

tests/tasks/setup/kit-cluster/default.yaml
```diff
@@ -81,25 +84,63 @@ spec:
...
+      ls -larth $(workspaces.config.path)
+      cat "$(workspaces.config.path)/dp$1"."yaml"
+      kubectl create -f "$(workspaces.config.path)/dp$1"."yaml"
+      }
+      for i in $(seq 1 $asgs)
+      do
+      #max number of nodes KIT operators allows per ASG
+      create_dp $i 1000 
+      done
+      remaining_nodes=$((nodes%1000))
```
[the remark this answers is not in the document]
hakuna-matatah: i don't think so. Because the range is 1 to 0, it will never get into loop.

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: With the current state of things with etcd and KIT cluster, all of the nodes becoming `ready` no matter how much ever time we wait is not a reality (very slim possibility).

prateekgogia: then why do we even wait for some nodes to be ready? We should narrow down whats the issue and fix it.

tests/tasks/setup/kit-cluster/default.yaml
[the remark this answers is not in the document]
prateekgogia: yeah you are right, I was mistaken by the loop seq.

tests/tasks/generators/clusterloader/load.yaml
prateekgogia: Sorry not clear which line uploads the `overrides.yaml` to S3? Is it line # 84 `aws s3 cp . s3://$(params.results-bucket)/ --recursive` ?

hakuna-matatah: Yes, upload step.

tests/tasks/setup/kit-cluster/default.yaml
hakuna-matatah: https://github.com/awslabs/kubernetes-iteration-toolkit/issues/86

prateekgogia (approved):
LGTM!

prateekgogia (approved):

PR #111, opened 2022-01-12
Moved from ASG to Fleet for faster devloop

ellistarn:
Issue #, if available:

Description of changes:


substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -20,68 +20,109 @@ import (
...
-			Value:             aws.String(substrate.Name),
-			PropagateAtLaunch: aws.Bool(true),
-		}, {
-			Key:               aws.String("Name"),
-			Value:             discovery.Name(substrate),
-			PropagateAtLaunch: aws.Bool(true),
-		}},
-	}); err != nil {
-		if strings.Contains(err.Error(), "Invalid IAM Instance Profile name") { // Instance Profile can take several seconds to propagate
+	instancesOutput, err := i.EC2.DescribeInstancesWithContext(ctx, &ec2.DescribeInstancesInput{})
```
prateekgogia, line 40: I think we should either use some filtering here or pagination because if an account/region has a large number of instances this can be an expensive call listing all the instances and it might timeout.

substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -20,68 +20,109 @@ import (
...
+			LaunchTemplateSpecification: &ec2.FleetLaunchTemplateSpecificationRequest{
+				LaunchTemplateName: discovery.Name(substrate),
+				Version:            substrate.Status.Cluster.LaunchTemplateVersion,
+			}},
+		},
+		TargetCapacitySpecification: &ec2.TargetCapacitySpecificationRequest{
+			DefaultTargetCapacityType: aws.String(ec2.DefaultTargetCapacityTypeOnDemand),
+			TotalTargetCapacity:       aws.Int64(1),
+		},
+		TagSpecifications: discovery.Tags(substrate, ec2.ResourceTypeInstance, discovery.Name(substrate)),
```
prateekgogia, line 73: Does fleet automatically adds the launch template version tag `"aws:ec2launchtemplate:version"` to the instance?

substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -20,68 +20,109 @@ import (
...
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/awslabs/kit/substrate/pkg/apis/v1alpha1"
 	"github.com/awslabs/kit/substrate/pkg/utils/discovery"
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 type Instance struct {
+	EC2         *ec2.EC2
 	AutoScaling *autoscaling.AutoScaling
```
prateekgogia, line 33: Can we remove this field as well now?

substrate/pkg/controller/substrate/cluster/launchtemplate.go
```diff
@@ -84,10 +85,21 @@ func (l *LaunchTemplate) Create(ctx context.Context, substrate *v1alpha1.Substra
 	} else {
 		logging.FromContext(ctx).Infof("Created launch template %s", aws.StringValue(discovery.Name(substrate)))
 	}
-	if _, err := l.EC2.CreateLaunchTemplateVersionWithContext(ctx, &ec2.CreateLaunchTemplateVersionInput{LaunchTemplateName: discovery.Name(substrate), LaunchTemplateData: launchTemplateData}); err != nil {
+	// Only update the launch template if it's changed
+	hash, err := hashstructure.Hash(launchTemplateData, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: true})
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("hashing launch template, %w", err)
+	}
+	launchTemplateVersionOutput, err := l.EC2.CreateLaunchTemplateVersionWithContext(ctx, &ec2.CreateLaunchTemplateVersionInput{
```
prateekgogia, line 93: Making sure I understand the flow, so if someone is using the CLI for the first time, LT version will start with 2?
Version 1 will be created when LT gets created and the version will be revised by `CreateLaunchTemplateVersionWithContext` to 2?

prateekgogia (changes requested):

substrate/pkg/controller/substrate/cluster/instance.go
ellistarn: Nice catch!!

substrate/pkg/controller/substrate/cluster/instance.go
ellistarn: Yep!

substrate/pkg/controller/substrate/cluster/launchtemplate.go
ellistarn: Yeah it's a bit odd -- agreed.

substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -20,68 +20,106 @@ import (
...
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("creating fleet, %w", err)
+	}
+	for _, err := range createFleetOutput.Errors {
+		if strings.Contains(aws.StringValue(err.ErrorMessage), "Invalid IAM Instance Profile name") {
 			return reconcile.Result{Requeue: true}, nil
 		}
-		if err.(awserr.Error).Code() != autoscaling.ErrCodeAlreadyExistsFault {
-			return reconcile.Result{}, fmt.Errorf("creating autoscaling group, %w", err)
+	}
```
prateekgogia, line 81: There can be other errors also in `createFleetOutput.Errors`? Looks like we should be returning an error if any other error exist in `createFleetOutput.Errors`?

prateekgogia (changes requested):

substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -78,6 +78,7 @@ func (i *Instance) Create(ctx context.Context, substrate *v1alpha1.Substrate) (r
 		if strings.Contains(aws.StringValue(err.ErrorMessage), "Invalid IAM Instance Profile name") {
 			return reconcile.Result{Requeue: true}, nil
 		}
+		return reconcile.Result{}, fmt.Errorf("creating fleet", aws.StringValue(err.ErrorMessage))
```
prateekgogia, line 81: If there are other/more errors we want to return an error, but if we know its `Invalid IAM Instance Profile name` we want to requeue.
I guess its fine, because if the first error is not `Invalid IAM Instance Profile name` we will return an error.

prateekgogia (approved):
Nicely done!

PR #113, opened 2022-01-14
Adds cluster config generation with kubeadm

prateekgogia:
Issue #, if available:
Description of changes:
- Creates an S3 bucket
- Generate kubeadm config for master and etcd components
- Uploads these config to S3 bucket
- Configure user-data to pull from S3 bucket
- Configure kubelet config for systemd


substrate/pkg/controller/substrate/infrastructure/subnets.go
```diff
@@ -46,6 +46,9 @@ func (s *Subnets) Create(ctx context.Context, substrate *v1alpha1.Substrate) (re
 		return reconcile.Result{}, err
 	}
 	for _, subnet := range subnets {
+		if subnet == nil { // we can run into a case when ctx is canceled, errs and subnets are all nil
```
ellistarn, line 49: If ctx is cancelled, don't you have an error value caught on line 45?

substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -78,7 +78,7 @@ func (i *Instance) Create(ctx context.Context, substrate *v1alpha1.Substrate) (r
 		if strings.Contains(aws.StringValue(err.ErrorMessage), "Invalid IAM Instance Profile name") {
 			return reconcile.Result{Requeue: true}, nil
 		}
-		return reconcile.Result{}, fmt.Errorf("creating fleet", aws.StringValue(err.ErrorMessage))
+		return reconcile.Result{}, fmt.Errorf("creating fleet %v", aws.StringValue(err.ErrorMessage))
```
ellistarn, line 81: Cool -- looks like %v has cleaner output than %s for nil values: https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w. I should start using this.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	return nil
+}
+
+func (c *Config) ensureS3Bucket(ctx context.Context, substrate *v1alpha1.Substrate) error {
+	if _, err := c.S3.CreateBucket(&s3.CreateBucketInput{
+		Bucket: discovery.Name(substrate),
+		CreateBucketConfiguration: &s3.CreateBucketConfiguration{
+			LocationConstraint: c.S3.Config.Region,
+		},
```
ellistarn, line 133:
```suggestion
		CreateBucketConfiguration: &s3.CreateBucketConfiguration{LocationConstraint: c.S3.Config.Region},
```

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	}); err != nil && err.(awserr.Error).Code() != s3.ErrCodeBucketAlreadyOwnedByYou {
+		return fmt.Errorf("creating S3 bucket, %w", err)
+	}
+	logging.FromContext(ctx).Infof("Ensured s3 bucket %s", aws.StringValue(discovery.Name(substrate)))
```
ellistarn, line 137: Can we differentiate created vs found? This log line should say created. Inside the error statement, you should log "Found". There are a ton of examples of this.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	return nil
+}
+
+func (c *Config) uploadDirectories(ctx context.Context, substrate *v1alpha1.Substrate) error {
```
ellistarn, line 141: This method is really a one liner. Not worth creating a helper function.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	certsphase "k8s.io/kubernetes/cmd/kubeadm/app/phases/certs"
+	"k8s.io/kubernetes/cmd/kubeadm/app/phases/controlplane"
+	etcdphase "k8s.io/kubernetes/cmd/kubeadm/app/phases/etcd"
+	configutil "k8s.io/kubernetes/cmd/kubeadm/app/util/config"
+	"knative.dev/pkg/logging"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	clusterCertsBasePath = "/tmp/"
```
ellistarn, line 41: Can we use https://pkg.go.dev/io/ioutil#TempDir

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	return true && d.Err() == nil
+}
+
+// Err returns error of DirectoryIterator
+func (d *DirectoryIterator) Err() error {
+	return d.err
+}
+
+// UploadObject uploads a file
+func (d *DirectoryIterator) UploadObject() s3manager.BatchUploadObject {
```
ellistarn, line 247: Does this upload in parallel?

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	f := d.next.f
+	return s3manager.BatchUploadObject{
+		Object: &s3manager.UploadInput{
+			Bucket: &d.bucket,
+			Key:    &d.next.path,
+			Body:   f,
+		},
+		After: func() error {
+			return f.Close()
+		},
```
ellistarn, line 257:
```suggestion
		After: f.Close,
```

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+// Err returns error of DirectoryIterator
+func (d *DirectoryIterator) Err() error {
+	return d.err
+}
+
+// UploadObject uploads a file
+func (d *DirectoryIterator) UploadObject() s3manager.BatchUploadObject {
+	f := d.next.f
+	return s3manager.BatchUploadObject{
+		Object: &s3manager.UploadInput{
```
ellistarn, line 250: consider inlining, since it's so short.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// Err returns error of DirectoryIterator
+func (d *DirectoryIterator) Err() error {
+	return d.err
+}
+
+// UploadObject uploads a file
+func (d *DirectoryIterator) UploadObject() s3manager.BatchUploadObject {
+	f := d.next.f
```
ellistarn, line 248: why can't this be inlined?

substrate/pkg/controller/substrate/cluster/config.go
ellistarn: >The SDK enables you to pass custom iterators to the new batched operations. For example, if we want to upload a directory, none of the default iterators do this easily. The following example shows how to implement a custom iterator that uploads a directory to S3.

Man -- instead of writing docs, why not write a directory iterator :O

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	f, err := os.Open(d.filePaths[0])
+	d.err = err
+	d.next.f = f
+	d.next.path = d.filePaths[0]
+	d.filePaths = d.filePaths[1:]
+	return true && d.Err() == nil
+}
+
+// Err returns error of DirectoryIterator
+func (d *DirectoryIterator) Err() error {
```
ellistarn, line 242: This is kinda wild. I see you're just copying the docs, but this impl is pretty weak. We can leave as is for now.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	logging.FromContext(ctx).Infof("Ensured s3 bucket %s", aws.StringValue(discovery.Name(substrate)))
+	return nil
+}
+
+func (c *Config) uploadDirectories(ctx context.Context, substrate *v1alpha1.Substrate) error {
+	if err := c.S3Uploader.UploadWithIterator(ctx, NewDirectoryIterator(
+		aws.StringValue(discovery.Name(substrate)), localConfigPath(substrate))); err != nil {
+		return fmt.Errorf("uploading to S3 %w", err)
+	}
+	logging.FromContext(ctx).Infof("Uploaded configs from %q to bucket %q",
```
ellistarn, line 146:
```suggestion
	logging.FromContext(ctx).Infof("Uploaded cluster configuration to s3://%q",
```

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+		localConfigPath(substrate), aws.StringValue(discovery.Name(substrate)))
```
ellistarn, line 147: No need to print out the tempdir, I think.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+	return nil
+}
+
+func localConfigPath(substrate *v1alpha1.Substrate) string {
```
ellistarn, line 151: You can probably get away w/o this helper if you don't log it.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+		err := controlplane.CreateStaticPodFiles(path.Join(clusterCertsBasePath, substrate.Name, clusterManifestPath), "",
+			&cfg.ClusterConfiguration, &cfg.LocalAPIEndpoint, false, componentName)
+		if err != nil {
+			return fmt.Errorf("creating static pod file for %v, %w", componentName, err)
+		}
+	}
+	return nil
+}
+
+func (c *Config) ensureS3Bucket(ctx context.Context, substrate *v1alpha1.Substrate) error {
```
ellistarn, line 128:
```suggestion
func (c *Config) ensureBucket(ctx context.Context, substrate *v1alpha1.Substrate) error {
```

substrate/pkg/controller/substrate/cluster/config.go
ellistarn: Consider collapsing all of these calls into oneliners if theyre <120 lines.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,259 @@
...
+		localConfigPath(substrate), aws.StringValue(discovery.Name(substrate)))
+	return nil
+}
+
+func localConfigPath(substrate *v1alpha1.Substrate) string {
+	return path.Join(clusterCertsBasePath, substrate.Name)
+}
+
+func defaultClusterConfig(substrate *v1alpha1.Substrate) (*kubeadm.InitConfiguration, error) {
+	defaultStaticConfig, err := configutil.DefaultedStaticInitConfiguration()
```
ellistarn, line 156: Might be nice to runtime.Must() here. Hard to imagine this can actually error at runtime.

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: IIUC, this will create a temp dir with random strings as suffix, we want to discover existing certs and configs atleast locally if someone runs the cli again to create same substrate cluster.

substrate/pkg/controller/substrate/infrastructure/subnets.go
prateekgogia: No I ran into this, err was nil and because we have subnets slice with nils defined on line 40, we enter this for loop.

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: yeah I wanted something quick and that worked.
Later I found I also used this same logic in the earlier implementation Lol.

substrate/pkg/controller/substrate/cluster/config.go
ellistarn: Certs probably shouldn't change. Kubernete configs need to change.

ellistarn: You're also not guaranteed that the host os even has a /tmp dir, which the go library has mechanisms to handle.

substrate/pkg/controller/substrate/infrastructure/subnets.go
ellistarn: There's gotta be a bug somewhere else.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -0,0 +1,305 @@
...
+		aws.StringValue(discovery.Name(substrate)), path.Join(clusterCertsBasePath, aws.StringValue(discovery.Name(substrate))))); err != nil {
+		return reconcile.Result{}, fmt.Errorf("uploading to S3 %w", err)
+	}
+	logging.FromContext(ctx).Infof("Uploaded cluster configuration to s3://%s", aws.StringValue(discovery.Name(substrate)))
+	return reconcile.Result{}, nil
+}
+
+func (c *Config) Delete(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	// delete the s3 bucket
+	if err := s3manager.NewBatchDeleteWithClient(c.S3).Delete(ctx, s3manager.NewDeleteListIterator(
```
ellistarn, line 96: I don't think you need to delete the files. Just delete the bucket.

prateekgogia: It fails to delete the bucket if it still has files in it, I will check if there is a `--force` kind of option

ellistarn: No!?

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: lets discuss about this, I am probably missing some context here

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: Looked at this there is no force option in the request

substrate/pkg/controller/substrate/cluster/config.go
ellistarn: Can you cut an issue?

ellistarn (approved):

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: https://github.com/awslabs/kubernetes-iteration-toolkit/issues/119

Issue #87, opened 2021-12-20
[operator] Ability to specify volume type and IOPS through CRD spec for etcd


hakuna-matatah:
- storage is GP2 for etcd with default IOPS KIT operator gives is 100 which is far from the EKS default we have today. No wonder, we run into slow disk issues on etcd. We need ability to mention this from KIT CRD spec.

felix-zhe-huang:
You can assign me.

felix-zhe-huang:
It looks like Karpenter does not support declaration of EBS volume type and IOPS yet. Will summit a PR to karpenter to enable this.
Meanwhile, it means that this task may take longer than expected.

PR #124, opened 2022-01-27
Add support of AWS-EBS-CSI driver to specify EBS IOPS

felix-zhe-huang:
Issue #, if available:
#87 
Description of changes:
- Change the operator API to support storage spec.
- Add controlplane and storageclass examples.
- Need AWS-EBS-CSI driver.


operator/pkg/awsprovider/launchtemplate/reconciler.go
```diff
@@ -95,7 +95,7 @@ func (c *Controller) createLaunchTemplate(ctx context.Context, dataplane *v1alph
 	if err != nil {
 		return fmt.Errorf("getting security group for control plane nodes, %w", err)
 	}
-	clusterEndpoint, err := master.GetClusterEndpoint(ctx, c.kubeclient, types.NamespacedName{dataplane.Namespace, dataplane.Spec.ClusterName})
+	clusterEndpoint, err := master.GetClusterEndpoint(ctx, c.kubeclient, types.NamespacedName{Namespace: dataplane.Namespace, Name: dataplane.Spec.ClusterName})
```
ellistarn, line 98: try `client.ObjectKeyFromObject`

operator/pkg/controllers/etcd/statefulset.go
```diff
@@ -54,6 +54,18 @@ func (c *Controller) reconcileStatefulSet(ctx context.Context, controlPlane *v1a
 				},
 				Spec: etcdSpec,
 			},
+			VolumeClaimTemplates: []v1.PersistentVolumeClaim{{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "etcd-data",
+				},
+				Spec: v1.PersistentVolumeClaimSpec{
```
ellistarn, line 61: Instead of just exposing StorageClassName and Size, what if we put the entire PersistentVolumeClaimSpec inside the ETCDSpec? Then you could say ETCD.Storage.StorageClassName, ETCD.Storage.Resources or potentially, if inlined, ETCD.StorageClassName and ETCD.Resources (may be a touch confusing with ETCD.Containers[0].Resources)

operator/pkg/controllers/master/endpoint.go
```diff
@@ -58,7 +58,7 @@ func (c *Controller) getClusterEndpoint(ctx context.Context, nn types.Namespaced
 
 func GetClusterEndpoint(ctx context.Context, client client.Client, nn types.NamespacedName) (string, error) {
 	svc := &v1.Service{}
-	if err := client.Get(ctx, types.NamespacedName{nn.Namespace, ServiceNameFor(nn.Name)}, svc); err != nil {
+	if err := client.Get(ctx, types.NamespacedName{Namespace: nn.Namespace, Name: ServiceNameFor(nn.Name)}, svc); err != nil {
```
ellistarn, line 61: same

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
```diff
@@ -44,7 +45,17 @@ type ControlPlaneList struct {
...
+}
+
+type StorageSpec struct {
+	StorageClassName string            `json:"storageClassName,omitempty"`
+	Size             resource.Quantity `json:"size,omitempty"`
+}
+
+type Etcd struct {
+	Component   `json:",inline"`
+	StorageSpec StorageSpec `json:"storageSpec,omitempty"`
```
ellistarn, line 58: This violates kubernetes patterns to call it `StorageSpec`. Idiomatically, it would be `Storage`

ellistarn: Otherwise you end up with a lot of stuttering spec.etcdspec.storagespec.

operator/docs/examples/storageclass.yaml
```diff
@@ -0,0 +1,10 @@
+kind: StorageClass
```
ellistarn, line 1: Thoughts on using a default storage class so we don't need to specify in most cases? 
https://kubernetes.io/docs/concepts/storage/storage-classes/#the-storageclass-resource

ellistarn (commented):
Nice work!

operator/pkg/controllers/etcd/statefulset.go
felix-zhe-huang: That is exactly what I want to discuss with the team. To me if we are committed to use the AWS-EBS-CSI driver to provision the EBS, then we should definitely expose the entire PersistentVolumeClaimSpec. In addition, we should also dynamically provision storageclass on the fly so that we can specify storagetype and IOPS in the controlplane spec.

On the other hand, I do see that Karpenter can create customized launchtemplate. I need to think a bit more about how to keep Karpenter AWS agnostic.

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
felix-zhe-huang: yes you are right. What was I thinking.

operator/docs/examples/storageclass.yaml
felix-zhe-huang: You are 100% right. We definitely need a default storageclass. So if no storage class is defined in the controlplane spec, the default will be used. We need to specify a high IOPS for the default (currently 100 IOPS is like a drop in the bucket).

So to continue the discussion, I was thinking that if we use AWS-EBS-CSI driver, KIT should be able to create different storageclass for different tests so that we can also scale test the impact of storage (e.g., IOPS, size, type). This should be optional in the contorlplane spec.

operator/pkg/controllers/etcd/statefulset.go
ellistarn: There are many kinds of PVC. It's not tightly coupled to EBS. I expect that KIT will always require some PVC for etc storage (thought it may be RAMFS, EBS, XFS, etc).

felix-zhe-huang:
The new commit contains the refactored code that use vloumeclaimtemplate in the control plane CRDs. I tested and the default values work beautifully. However, I am encountering a bug that when a new volumeclaimtemplate spec is provided in the control plane CRD, the controlplane object created is missing the name (etcd.volumeClaimTemplates.metadata.name). The name is used to index the volumeclaim templates. As a result, this version does not support customized volumeclaim template yet. The attachment is a controlplane CRD that can recreate the problem.
[test_contrloplane.txt](https://github.com/awslabs/kubernetes-iteration-toolkit/files/7962468/test_contrloplane.txt)

operator/charts/kit-operator/templates/controller/storageclass_io1.yaml
```diff
@@ -0,0 +1,10 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: kit-io1
```
ellistarn, line 4: Any reason we prefix these with KIT?

operator/pkg/controllers/etcd/statefulset.go
```diff
@@ -32,10 +32,17 @@ func (c *Controller) reconcileStatefulSet(ctx context.Context, controlPlane *v1a
...
+	podSpec, err := patch.PodSpec(podSpecFor(controlPlane), controlPlane.Spec.Etcd.Spec)
 	if err != nil {
 		return fmt.Errorf("failed to patch pod spec, %w", err)
 	}
+	// Generate the default volume claim template spec for the given control plane, if user has
+	// provided custom config for the etcd volume template spec, patch this user
+	// provided config to the default spec
+	volumeClaimSpec, err := patch.VolumeClaimTemplateSpec(DefaultVolumeClaimTemplateSpec(), controlPlane.Spec.Etcd.VolumeClaimTemplates)
+	if err != nil {
+		return fmt.Errorf("failed to patch voluemclaim template spec, %w", err)
```
ellistarn, line 44: We know what these types are and control the input/output. Can we write them in a way that doesn't require expressive error handling? We can instead panic, which will only happen due to a programmer error.

operator/pkg/controllers/etcd/volumeclaimtemplates.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+		Spec: v1.PersistentVolumeClaimSpec{
+			AccessModes:      []v1.PersistentVolumeAccessMode{"ReadWriteOnce"},
+			StorageClassName: &storageClassName,
+			Resources: v1.ResourceRequirements{
+				Requests: v1.ResourceList{"storage": resource.MustParse("40Gi")},
+			},
+		},
+	}}
+
```
ellistarn, line 37:
```suggestion
```

operator/pkg/controllers/etcd/volumeclaimtemplates.go
```diff
@@ -0,0 +1,38 @@
...
+func DefaultVolumeClaimTemplateSpec() []v1.PersistentVolumeClaim {
+	storageClassName := "kit-gp3"
+	return []v1.PersistentVolumeClaim{{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "etcd-data",
+		},
+		Spec: v1.PersistentVolumeClaimSpec{
+			AccessModes:      []v1.PersistentVolumeAccessMode{"ReadWriteOnce"},
+			StorageClassName: &storageClassName,
```
ellistarn, line 31: try the knative library ptr for things like this. `ptr.String("kit-gp3")`. I  developed the FooValue() variants of this, including ptr.StringValue(), too.

operator/pkg/controllers/master/certificates.go
```diff
@@ -32,7 +32,7 @@ import (
 const (
 	rootCACommonName       = "kubernetes"
 	frontProxyCACommonName = "front-proxy-ca"
-	kubeAdminName          = "kubernetes-admin"
+	//kubeAdminName          = "kubernetes-admin"
```
ellistarn, line 35: delete

operator/pkg/utils/patch/patch.go
```diff
@@ -23,6 +23,47 @@ import (
...
 
+// VolumeClaimTemplateSpec returns the merged VolumeClaimTemplate spes
+func VolumeClaimTemplateSpec(defaultSpec, patch []v1.PersistentVolumeClaim) ([]v1.PersistentVolumeClaim, error) {
+	if len(patch) == 0 {
+		return defaultSpec, nil
+	}
+	defaults, patches := map[string]v1.PersistentVolumeClaim{}, map[string]v1.PersistentVolumeClaim{}
+	for _, spec := range defaultSpec {
+		if _, ok := defaults[spec.Name]; ok {
+			return nil, fmt.Errorf("duplicate PersistentVolumeClaim detected, Name: %s", spec.ObjectMeta.Name)
```
ellistarn, line 34: Now that I see this, maybe hard to avoid erroring.

felix-zhe-huang:
OK I did a long night debugging and here is my theory. The object metadata of volumeclaimtemplate is overridden by the API server maybe to update the creation timestamp. And the CRD API says that the metadata is an object and does not expand further. So it must be replaced as a whole and hence the name is gone.

I don't see any workaround since we don't control  volumeclaimtemplate. I have update the code to expose just the presistentvolumeclaim spec. We lose the capability to specify multiple templates but at least now the code works.

I think we should move forward and revisit this later when we see use cases that requires multiple PVC templates.

operator/charts/kit-operator/templates/controller/storageclass_io1.yaml
felix-zhe-huang: in my envionment EKS already created a class called gp2. It seems that naming the classes using the storage type is a popular idea. To make sure that we don't have a collision, I give it a prefix "kit-".

operator/pkg/utils/patch/patch.go
felix-zhe-huang: Yeah, when we need to process arbitrary user input, things tense to go wrong by a lot. Can't really avoid it.

operator/pkg/controllers/controlplane/suite_test.go
```diff
@@ -76,7 +76,7 @@ var _ = Describe("ControlPlane", func() {
 			Namespace: "default",
 		}, Spec: v1alpha1.ControlPlaneSpec{
 			Master: v1alpha1.MasterSpec{APIServer: &v1alpha1.Component{Replicas: 1}},
-			Etcd:   &v1alpha1.Component{}}}
+			Etcd:   v1alpha1.Etcd{Component: v1alpha1.Component{Replicas: 3, Spec: nil}, PersistentVolumeClaimSpec: nil}}}
```
prateekgogia, line 79: we can skip adding `Spec` and `PersistentVolumeClaimSpec` fields as they are nil by default?

operator/pkg/controllers/etcd/persistentvolumeclaimspec.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func DefaultPersistentVolumeClaimSpec() *v1.PersistentVolumeClaimSpec {
+	storageClassName := "kit-gp3"
+	return &v1.PersistentVolumeClaimSpec{
+		AccessModes:      []v1.PersistentVolumeAccessMode{"ReadWriteOnce"},
```
prateekgogia, line 25:
```suggestion
		AccessModes:      []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
```

operator/docs/examples/controlplane_iops.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: kit.k8s.sh/v1alpha1
+kind: ControlPlane
+metadata:
+  name: guest # Desired Cluster Name
+spec:
+  kubernetesVersion: "1.21"
+  etcd:
+    storageSpec:
```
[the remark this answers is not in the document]
felix-zhe-huang: Good catch! I have updated the example how to override the parameters.
But the default KIT CRD will work out of the box.

operator/pkg/controllers/etcd/persistentvolumeclaimspec.go
```diff
@@ -0,0 +1,32 @@
...
+		StorageClassName: &storageClassName,
```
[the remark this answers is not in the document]
felix-zhe-huang: done

operator/pkg/controllers/etcd/statefulset.go
```diff
@@ -52,8 +60,14 @@ func (c *Controller) reconcileStatefulSet(ctx context.Context, controlPlane *v1a
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: labelsFor(controlPlane.ClusterName()),
 				},
-				Spec: etcdSpec,
+				Spec: podSpec,
 			},
+			VolumeClaimTemplates: []v1.PersistentVolumeClaim{{
```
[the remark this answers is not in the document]
felix-zhe-huang: done

operator/pkg/utils/patch/patch.go
```diff
@@ -23,6 +23,23 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
+// VolumeClaimTemplateSpec returns the merged VolumeClaimTemplate spes
+func PersistentVolumeClaimSpec(defaultSpec, patch *v1.PersistentVolumeClaimSpec) (v1.PersistentVolumeClaimSpec, error) {
+	if patch == nil {
+		return *defaultSpec, nil
+	}
+	merged, err := mergePatch(defaultSpec, patch, v1.PersistentVolumeClaimSpec{})
```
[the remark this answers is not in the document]
felix-zhe-huang: IIUC, sometime we do want to introduce special logic in merging (e.g., PodSpec has this mergeContainerArgs that help merging the arguement instead of replacing one by other). A generic wrapper may not be able to do so without some extra complexity. Maybe when we see more cases when KIT need to expose specs, we can have a better understanding how to create a generic wrapper?

operator/pkg/controllers/controlplane/suite_test.go
felix-zhe-huang: good point. Done

operator/pkg/controllers/etcd/persistentvolumeclaimspec.go
felix-zhe-huang: done

operator/pkg/utils/patch/patch.go
```diff
@@ -23,6 +23,23 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
+// VolumeClaimTemplateSpec returns the merged VolumeClaimTemplate spes
+func PersistentVolumeClaimSpec(defaultSpec, patch *v1.PersistentVolumeClaimSpec) (v1.PersistentVolumeClaimSpec, error) {
```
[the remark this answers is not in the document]
felix-zhe-huang: I don't have a good answer. @prateekgogia I saw that the PodSpec patching is also using this pattern. Is there any reason that we can't return a pointer as well?

operator/pkg/controllers/etcd/persistentvolumeclaimspec.go
```diff
@@ -0,0 +1,32 @@
...
+
+func DefaultPersistentVolumeClaimSpec() *v1.PersistentVolumeClaimSpec {
+	return &v1.PersistentVolumeClaimSpec{
+		AccessModes:      []v1.PersistentVolumeAccessMode{v1.ReadWriteOnce},
+		StorageClassName: ptr.String("kit-gp3"),
+		Resources: v1.ResourceRequirements{
+			Requests: v1.ResourceList{"storage": resource.MustParse("40Gi")},
+		},
+	}
+
```
ellistarn, line 31:
```suggestion
```

prateekgogia:
I have filled an issue for pending tasks https://github.com/awslabs/kubernetes-iteration-toolkit/issues/133
Happy to merge this as-is!

PR #127, opened 2022-02-02
Support for addons in substrate node

prateekgogia:
Issue #, if available:

Description of changes:
- Create addons for substrate kube-proxy and aws-iam-authenticator
- Add another package for aws-iam-authenticator specific common code in operator and substrate
- Remove sleep when doing `aws s3 sync` on substrate node
- Kubelet running on the substrate node will register itself to the API server on the same node.
- Some bug fixes


operator/pkg/controllers/master/authenticatorconfig.go
```diff
@@ -35,146 +28,10 @@ func (c *Controller) reconcileAuthenticatorConfig(ctx context.Context, controlPl
...
-		AWSAccountID: awsAccountID,
-		PrivateDNS:   "{{EC2PrivateDNSName}}",
-		SessionName:  "{{SessionNameRaw}}",
-	})
-	if err != nil {
-		return fmt.Errorf("generating authenticator config, %w", err)
-	}
-	return c.kubeClient.EnsurePatch(ctx, &v1.ConfigMap{}, configMap)
+	return c.awsIamAuthenticator.EnsureConfig(ctx, controlPlane.ClusterName(), controlPlane.Namespace,
+		iam.KitNodeRoleNameFor(controlPlane.ClusterName()), awsAccountID)
```
ellistarn, line 32: Potentially noop for this PR, but we should avoid using terms like KIT inside of KIT, as it makes renaming the project very had. I think iam.NodeRoleName(...) is plenty expressive.

operator/pkg/controllers/master/authenticatorconfig.go
```diff
@@ -35,146 +28,10 @@ func (c *Controller) reconcileAuthenticatorConfig(ctx context.Context, controlPl
...
-        - {{ .AWSAccountID }}
-`
-)
-func AuthenticatorDaemonSetName(clusterName string) string {
-	return fmt.Sprintf("%s-authenticator", clusterName)
-}
-
-func AuthenticatorConfigMapName(clusterName string) string {
-	return fmt.Sprintf("%s-auth-config", clusterName)
+	return c.awsIamAuthenticator.EnsureDaemonSet(ctx, controlPlane, APIServerLabels(controlPlane.ClusterName()))
```
ellistarn, line 36: Seems odd to wrap this function if these are straight passthroughs. Why can't the caller be aware and dedup this?

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kuberuntime "k8s.io/apimachinery/pkg/runtime"
+	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/pkg/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Controller struct {
```
ellistarn, line 35: Why are we defining a controller in utils? Shouldn't this be in controllers?

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+	KubeClient *kubeprovider.Client
```
ellistarn, line 36: why public?

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func New(KubeClient *kubeprovider.Client) *Controller {
```
ellistarn, line 39:
```suggestion
func New(kubeClient *kubeprovider.Client) *Controller {
```

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+
+func PodSpec(obj client.Object, opts ...Options) v1.PodSpec {
+	spec := podSpec(obj)
+	for _, opt := range opts {
+		spec = opt(spec)
+	}
+	return spec
+}
+
+func podSpec(obj client.Object) v1.PodSpec {
```
ellistarn, line 112: Why does this take an obj?

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+			VolumeSource: v1.VolumeSource{
+				HostPath: &v1.HostPathVolumeSource{
+					Path: "/var/aws-iam-authenticator/state/",
+				},
+			},
+		}},
+	}
+}
+
+func authenticatorLabels() map[string]string {
```
ellistarn, line 171: If this is static, why is it a function? Can we autogenerate these somehow, assuming all components generically have the label "component"

operator/pkg/utils/iamauthenticator/awsiamauthenticator.go
```diff
@@ -0,0 +1,215 @@
...
+	var buf bytes.Buffer
+	tmpl := template.Must(template.New("Text").Parse(strtmpl))
+	err := tmpl.Execute(&buf, obj)
+	if err != nil {
+		return nil, fmt.Errorf("error when executing template, %w", err)
+	}
+	return buf.Bytes(), nil
+}
+
+func AuthenticatorDaemonSetName(clusterName string) string {
```
ellistarn, line 209: You're already in the authenticator package, no need to stutter on naming. Any reason we can't unify the names of all these things? "cluster:component" or "etarndev:authenticator"

operator/pkg/utils/imageprovider/imageprovider.go
```diff
@@ -60,9 +60,9 @@ func CoreDNS() string {
 }
 
 func AWSIamAuthenticator() string {
```
ellistarn, line 62: I'm not sure how I feel about this image provider abstraction. It breaks the encapsulation of the components themselves (e.g. the actual authenticator), for seemingly little value.

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/awslabs/kit/substrate/pkg/controller/substrate/cluster"
+	proxyaddon "k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/proxy"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type KubeProxy struct {
+}
+
+func (k *KubeProxy) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	client, err := KubeClientFor(ctx, substrate)
```
ellistarn, line 32: Much cleaner to follow the existing pattern and check for prerequisites (e.g. substrate.endpoint) before making calls like this.

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+package addons
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/awslabs/kit/substrate/pkg/apis/v1alpha1"
+	"github.com/awslabs/kit/substrate/pkg/controller/substrate/cluster"
+	proxyaddon "k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/proxy"
```
ellistarn, line 24: why does this need an alias?

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (k *KubeProxy) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	client, err := KubeClientFor(ctx, substrate)
+	if errors.Is(err, ErrWaitingForSubstrateEndpoint) {
+		return reconcile.Result{Requeue: true}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+	// deploy kube proxy addon
```
ellistarn, line 38:
```suggestion
```

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+	proxyaddon "k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/proxy"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type KubeProxy struct {
+}
+
+func (k *KubeProxy) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	client, err := KubeClientFor(ctx, substrate)
+	if errors.Is(err, ErrWaitingForSubstrateEndpoint) {
```
ellistarn, line 33: I'm not sure you need this abstraction `ErrWaitingForSubstrateEndpoint` at all if you follow the recommendation of the comment above.

substrate/pkg/controller/substrate/cluster/addons/rbac.go
```diff
@@ -0,0 +1,54 @@
...
+
+package addons
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/awslabs/kit/substrate/pkg/apis/v1alpha1"
+	nodebootstraptokenphase "k8s.io/kubernetes/cmd/kubeadm/app/phases/bootstraptoken/node"
```
ellistarn, line 23: I'd avoid aliasing unless you have a naming conflict.

substrate/pkg/controller/substrate/cluster/addons/rbac.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (r *RBAC) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	client, err := KubeClientFor(ctx, substrate)
+	if errors.Is(err, ErrWaitingForSubstrateEndpoint) {
+		return reconcile.Result{Requeue: true}, nil
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
+	// Create RBAC rules that makes the bootstrap tokens able to post CSRs
+	if err := nodebootstraptokenphase.AllowBootstrapTokensToPostCSRs(client); err != nil {
```
ellistarn, line 38: Seems like you're somewhat forced into this pattern by kubeadm. Do you know how long each of these calls take?

substrate/pkg/controller/substrate/cluster/addons/utils.go
```diff
@@ -0,0 +1,78 @@
...
+
+const (
+	kubeAdminFilePath = "etc/kubernetes/admin.conf"
+)
+
+func KubeClientFor(ctx context.Context, substrate *v1alpha1.Substrate) (*kubernetes.Clientset, error) {
+	if substrate.Status.Cluster.Address == nil {
+		return nil, ErrWaitingForSubstrateEndpoint
+	}
+	// check if the kube-admin file for the cluster exists
```
ellistarn, line 47: Will this race? Is there a way that you can inject a client as a dependency that can be shared between callers?

substrate/pkg/controller/substrate/controller.go
```diff
@@ -56,7 +58,10 @@ func NewController(ctx context.Context) *Controller {
 			&cluster.LaunchTemplate{EC2: EC2, SSM: ssm.New(session), Region: session.Config.Region},
 			&cluster.InstanceProfile{IAM: IAM},
 			&cluster.Instance{EC2: EC2},
-			&cluster.Config{S3: s3.New(session), S3Uploader: s3manager.NewUploader(session)},
+			&cluster.Config{S3: s3.New(session), StsClient: sts.New(session), S3Uploader: s3manager.NewUploader(session)},
+			// Addons installed on substrate cluster
```
ellistarn, line 62: I think this is pretty clear due to our package naming convention
```suggestion
```

substrate/pkg/controller/substrate/controller.go
```diff
@@ -56,7 +58,10 @@ func NewController(ctx context.Context) *Controller {
 			&cluster.LaunchTemplate{EC2: EC2, SSM: ssm.New(session), Region: session.Config.Region},
 			&cluster.InstanceProfile{IAM: IAM},
 			&cluster.Instance{EC2: EC2},
-			&cluster.Config{S3: s3.New(session), S3Uploader: s3manager.NewUploader(session)},
+			&cluster.Config{S3: s3.New(session), StsClient: sts.New(session), S3Uploader: s3manager.NewUploader(session)},
```
ellistarn, line 61:
```suggestion
			&cluster.Config{S3: s3.New(session), STS: sts.New(session), S3Uploader: s3manager.NewUploader(session)},
```

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -53,103 +53,169 @@ var PolicyDocument = aws.String(`{
...
 }`)
 
-var ManagedPolicies = []string{
+var ManagedPoliciesForSubstrateNode = []string{
 	"arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore",
+	"arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
+	"arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+}
+
+var ManagedPoliciesForNodeProvisionedByKarpenter = []string{
```
ellistarn, line 83: Consider inlining where this is used, since it's all statically defined.

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -53,103 +53,169 @@ var PolicyDocument = aws.String(`{
...
+				"ec2:DescribeAvailabilityZones",
+				"ssm:GetParameter"
 			],
 			"Resource": ["*"]
 		}
 	]
 }`)
 
-var ManagedPolicies = []string{
+var ManagedPoliciesForSubstrateNode = []string{
```
ellistarn, line 77: Similar

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -44,7 +44,7 @@ var AssumeRolePolicyDocument = aws.String(`{
 	]
 }`)
 
-var PolicyDocument = aws.String(`{
+var SubstrateNodePolicyDocument = aws.String(`{
```
ellistarn, line 47: Similar

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -53,103 +53,169 @@ var PolicyDocument = aws.String(`{
...
+}
+
+type roleInfo struct {
+	objectName      *string
+	policyDocument  *string
+	managedPolicies []string
 }
 
 func (i *InstanceProfile) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	for _, desired := range desiredRolesFor(substrate) {
```
ellistarn, line 98: Are we going to have more than two roles? I think the substrate has: control plane role, data plane role. I'm not sure we'll grow beyond this. Rather than a new type, struct indirection, and looping, it might be more straight forward to just explicitly create both of them.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -28,35 +28,43 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/aws/aws-sdk-go/service/sts"
+	iamauthenticatoraddon "github.com/awslabs/kit/operator/pkg/utils/iamauthenticator"
```
ellistarn, line 32: I think we should avoid this dependency if at all possible. Else we tightly couple the dependency closure (and k8s version upgrade) of these projects.

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
prateekgogia: The reason I added this abstraction was I didn't want to return 3 results like -
```
func KubeClientFor(ctx context.Context, substrate *v1alpha1.Substrate) (*kubernetes.Clientset, reconcile.Result, error) {
```
I just indicate the caller with `ErrWaitingForSubstrateEndpoint` to requeue or not when there is an error.

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
prateekgogia: This was to avoid some duplicate code, so that all addons (kube-proxy, rbac) are not doing the same check and its done at `KubeClientFor` level.

substrate/pkg/controller/substrate/cluster/addons/rbac.go
prateekgogia: Its less than a second.

substrate/pkg/controller/substrate/cluster/addons/utils.go
prateekgogia: > Will this race?

You mean race between controllers? Right now each controller for addon gets their own kubeclient.

>Is there a way that you can inject a client as a dependency to your other controllers that can be shared between callers?

I looked into this to be able to pass the same client to all controllers, but currently the way its designed is all the controllers are started in `substrate/controller.go` makes it tough to achieve. Other alternative can be have a wrapper which wraps all the addon controllers, and calls create only when kubeclient is created and can be shared between controllers

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: There is a lot of duplicate code like in this case which I was trying to reuse. I think long term we need to discuss if we want to extract out common code into a separate package

substrate/pkg/controller/substrate/cluster/instanceprofile.go
prateekgogia: There is one more role we will need to create for guest dataplane nodes.

operator/pkg/utils/imageprovider/imageprovider.go
prateekgogia: The reason I added this package was to have all the images defined at a common place and not spread them all over the code. So if we need to change a tag/version for a component we can just change in this one file.

operator/pkg/utils/iamauthenticator/iamauthenticator.go
```diff
@@ -0,0 +1,149 @@
+/*
```
ellistarn, line 1: operator/pkg/components/iamauthenticator.go

iamauthenticator.Pod(...) v1.PodTemplateSpec

In general, I think this is a decent pattern for all of our components (kcm/kcp/ksh).

prateekgogia:
Things remaining which I will be covering in the next PR-
- Generate `component` label for all objects created, deployments, authenticator etc.
- Refactor `EnsurePatch` method to not require object type in args.

operator/pkg/utils/imageprovider/imageprovider.go
```diff
@@ -60,9 +60,9 @@ func CoreDNS() string {
...
-	return "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-iam-authenticator:v0.5.3-amazonlinux-2"
+	return RepositoryName + "kubernetes-sigs/aws-iam-authenticator:v0.5.3-eks-1-21-8"
 }
 
 func BusyBox() string {
-	return "public.ecr.aws/runecast/busybox:1.33.1"
+	return "public.ecr.aws/docker/library/busybox:stable"
```
ellistarn, line 67: Thoughts on just making this a constant?

substrate/pkg/apis/v1alpha1/substrate.go
```diff
@@ -17,7 +17,14 @@ limitations under the License.
...
 
 import (
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
+)
+
+const (
+	// SubstrateReady means
+	SubstrateReady apis.ConditionType = "SubstrateReady"
```
ellistarn, line 27:
```suggestion
	Ready apis.ConditionType = "Ready"
```

substrate/pkg/apis/v1alpha1/substrate.go
```diff
@@ -50,3 +57,19 @@ type SubnetSpec struct {
 	CIDR   string
 	Public bool
 }
+
+func (s *Substrate) IsReady() bool {
```
ellistarn, line 61: You should use a living condition set

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+	if !substrate.IsReady() {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	client, err := kubeconfig.ClientSetFromFile(*substrate.Status.Cluster.KubeConfig)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("creating Kube client from admin config, %w", err)
+	}
+	config := cluster.DefaultClusterConfig(substrate)
+	if err := proxy.EnsureProxyAddon(&config.ClusterConfiguration, &config.LocalAPIEndpoint, client); err != nil {
+		return reconcile.Result{Requeue: true}, fmt.Errorf("deploying kube-proxy addon, %w", err)
```
ellistarn, line 41:
```suggestion
		return reconcile.Result{Requeue: true}, fmt.Errorf("ensuring kube-proxy addon, %w", err)
```

substrate/pkg/controller/substrate/cluster/addons/kubeproxy.go
```diff
@@ -0,0 +1,48 @@
...
+type KubeProxy struct {
+}
+
+func (k *KubeProxy) Create(ctx context.Context, substrate *v1alpha1.Substrate) (reconcile.Result, error) {
+	if !substrate.IsReady() {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	client, err := kubeconfig.ClientSetFromFile(*substrate.Status.Cluster.KubeConfig)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("creating Kube client from admin config, %w", err)
```
ellistarn, line 37:
```suggestion
		return reconcile.Result{}, fmt.Errorf("creating client, %w", err)
```

substrate/pkg/controller/substrate/cluster/addons/rbac.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := node.AllowBootstrapTokensToPostCSRs(client); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error allowing bootstrap tokens to post CSRs, %w", err)
+	}
+	// Create RBAC rules that makes the bootstrap tokens able to get their CSRs approved automatically
+	if err := node.AutoApproveNodeBootstrapTokens(client); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error auto-approving node bootstrap tokens, %w", err)
+	}
+	// Create/update RBAC rules that makes the nodes to rotate certificates and get their CSRs approved automatically
+	if err := node.AutoApproveNodeCertificateRotation(client); err != nil {
+		return reconcile.Result{}, fmt.Errorf("err AutoApproveNodeCertificateRotation, %w", err)
```
ellistarn, line 48: be consistent with these error messages.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -250,6 +273,47 @@ func defaultClusterConfig(substrate *v1alpha1.Substrate) *kubeadm.InitConfigurat
...
+		return fmt.Errorf("failed to create directory, %w", err)
+	}
+	if err := ioutil.WriteFile(path.Join(configDir, "config.yaml"), []byte(configMap.Data["config.yaml"]), 0644); err != nil {
+		return fmt.Errorf("writing authenticator config, %w", err)
+	}
+	return nil
+}
+
+func (c *Config) staticPodSpecForAuthenticator(ctx context.Context, substrate *v1alpha1.Substrate) error {
+	podTemplateSpec := iamauthenticator.PodSpec(func(spec v1.PodSpec) v1.PodSpec {
```
ellistarn, line 298: Shouldn't this be a pod template spec, so that you can modify everything you need in one go? Seems like you're getting in your own way on line 304.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -250,6 +273,47 @@ func defaultClusterConfig(substrate *v1alpha1.Substrate) *kubeadm.InitConfigurat
...
+		spec.Volumes = append(spec.Volumes, v1.Volume{Name: "config",
+			VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: authenticatorConfigDir}},
+		})
+		return spec
+	})
+	podTemplateSpec.ObjectMeta.Namespace = "kube-system"
+	serialized, err := kubeadmutil.MarshalToYaml(
```
ellistarn, line 305: It seems wrong to put this logic in the static pod spec method. This approach couples side effects to static generation. I think this logic probably should live in ensureAuthenticatorConfig.

ellistarn: Can you not do:

`kubeadmutil.MarshalToYaml(iamauthenticator.PodSpec(...))`

in `ensureAuthenticatorConfig`

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -31,7 +31,25 @@ type InstanceProfile struct {
 	IAM *iam.IAM
 }
 
-var AssumeRolePolicyDocument = aws.String(`{
+type roleInfo struct {
+	objectName      *string
+	policyDocument  *string
+	managedPolicies []string
```
ellistarn, line 37:
```suggestion
	policies []string
```

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -31,7 +31,25 @@ type InstanceProfile struct {
 	IAM *iam.IAM
 }
 
-var AssumeRolePolicyDocument = aws.String(`{
+type roleInfo struct {
+	objectName      *string
```
ellistarn, line 35:
```suggestion
	name      *string
```

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -31,7 +31,25 @@ type InstanceProfile struct {
...
+	policyDocument  *string
```
ellistarn, line 36:
```suggestion
	policy  *string
```

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -41,115 +59,154 @@ var AssumeRolePolicyDocument = aws.String(`{
...
 			return reconcile.Result{}, fmt.Errorf("deleting role, %w", err)
 		}
 	} else {
-		logging.FromContext(ctx).Infof("Deleted role %s", aws.StringValue(discovery.Name(substrate)))
+		logging.FromContext(ctx).Infof("Deleted role %s", aws.StringValue(resourceName))
 	}
 	return reconcile.Result{}, nil
 }
+
+func desiredRolesFor(substrate *v1alpha1.Substrate) []roleInfo {
```
ellistarn, line 164: scope this function to the instanceprofile controller

func (i *InstanceProfile) roles(substrate *v1alpha1.Substrate)

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -31,7 +31,25 @@ type InstanceProfile struct {
 	IAM *iam.IAM
 }
 
-var AssumeRolePolicyDocument = aws.String(`{
+type roleInfo struct {
```
ellistarn, line 34: Describing a variable name with _Info should trigger naming alarm bells in your head. What value does `Info` convey here? I'd probably just name this `role`, since this concept encapsulates a role in this package.

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -41,115 +59,154 @@ var AssumeRolePolicyDocument = aws.String(`{
...
 	}
 	// Policy
-	if _, err := i.IAM.PutRolePolicyWithContext(ctx, &iam.PutRolePolicyInput{RoleName: discovery.Name(substrate), PolicyName: discovery.Name(substrate), PolicyDocument: PolicyDocument}); err != nil {
-		return reconcile.Result{}, fmt.Errorf("adding policy to role, %w", err)
-	} else {
-		logging.FromContext(ctx).Infof("Created policy %s for %s", aws.StringValue(discovery.Name(substrate)), aws.StringValue(discovery.Name(substrate)))
+	if policyDocument != nil {
+		if _, err := i.IAM.PutRolePolicyWithContext(ctx, &iam.PutRolePolicyInput{RoleName: resourceName, PolicyName: resourceName, PolicyDocument: policyDocument}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("adding policy to role, %w", err)
+		} else {
```
ellistarn, line 74: You don't need this else statement, since the if before it returns.

substrate/pkg/controller/substrate/cluster/launchtemplate.go
```diff
@@ -58,7 +58,9 @@ func (l *LaunchTemplate) Create(ctx context.Context, substrate *v1alpha1.Substra
 		ImageId:            parameterOutput.Parameter.Value,
 		IamInstanceProfile: &ec2.LaunchTemplateIamInstanceProfileSpecificationRequest{Name: discovery.Name(substrate)},
 		Monitoring:         &ec2.LaunchTemplatesMonitoringRequest{Enabled: aws.Bool(true)},
-		SecurityGroupIds:   []*string{substrate.Status.SecurityGroupID},
+		SecurityGroupIds:   []*string{substrate.Status.Infrastructure.SecurityGroupID},
+		// aws s3 sync sometimes fails to sync small changes in a file, so we use --exact-timestamps
```
ellistarn, line 62: I'm not sure what this comment refers to.

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -250,6 +273,47 @@ func defaultClusterConfig(substrate *v1alpha1.Substrate) *kubeadm.InitConfigurat
...
 
+func (c *Config) ensureAuthenticatorConfig(ctx context.Context, substrate *v1alpha1.Substrate) error {
+	identity, err := c.STS.GetCallerIdentityWithContext(ctx, &sts.GetCallerIdentityInput{})
+	if err != nil {
+		return fmt.Errorf("getting caller identity, %w", err)
+	}
+	configMap, err := iamauthenticator.Config(ctx, substrate.Name, substrate.Namespace,
+		aws.StringValue(discovery.Name(substrate, tenantControlPlaneNodeRole)), aws.StringValue(identity.Account))
+	if err != nil {
+		return err
```
ellistarn, line 284: any reason this isn't decorated?

substrate/pkg/apis/v1alpha1/substrate.go
prateekgogia: I used in the next revision of the PR but I am not sure whats the benefit and even if I am using it the right way.

substrate/pkg/controller/substrate/cluster/instanceprofile.go
prateekgogia: Is there a benefit in doing this? I usually try to avoid having a receiver if not required.

substrate/pkg/controller/substrate/cluster/config.go
prateekgogia: > I think this logic probably should live in ensureAuthenticatorConfig

Do you mean `ensureAuthenticatorConfig` and `staticPodSpecForAuthenticator` should be one single function?
Also trying to understand as this only marshaling podSpec to a yaml, is this considered to be a static generation?

operator/pkg/components/iamauthenticator/iamauthenticator.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/awslabs/kit/operator/pkg/utils/imageprovider"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"knative.dev/pkg/ptr"
+)
+
+func Config(ctx context.Context, name, ns, instanceRole, accountID string) (*v1.ConfigMap, error) {
```
ellistarn, line 31: Optional, name it what it is:
```suggestion
func ConfigMap(ctx context.Context, name, ns, instanceRole, accountID string) (*v1.ConfigMap, error) {
```

operator/pkg/components/iamauthenticator/iamauthenticator.go
```diff
@@ -0,0 +1,157 @@
...
+// TODO move this to util. parseTemplate validates and parses passed as argument template
+func parseTemplate(strtmpl string, obj interface{}) ([]byte, error) {
+	var buf bytes.Buffer
+	if err := template.Must(template.New("Text").Parse(strtmpl)).Execute(&buf, obj); err != nil {
+		return nil, fmt.Errorf("error when executing template, %w", err)
+	}
+	return buf.Bytes(), nil
+}
+
+func AuthenticatorConfigMapName(clusterName string) string {
```
ellistarn, line 151: nit: stuttering.
```suggestion
func Name(clusterName string) string {
```

operator/pkg/utils/imageprovider/imageprovider.go
ellistarn: Willing to see how this goes.

substrate/pkg/apis/v1alpha1/substrate.go
ellistarn: I think this isn't the right way, but I'd have to play with it and/or dig into tekton.

ellistarn (approved):

ellistarn (approved):
Nice work, overall in really great shape. Just a couple minor things that we can address later.

prateekgogia:
Will address the minor things in next PR.

Issue #81, opened 2021-11-24
[operator] Configure health-checks for KIT control-plane components


hakuna-matatah:
Health-checks are required for KIT control-plane components to not mask issues.

PR #128, opened 2022-02-03
Adding liveness probes to control plane components

natherz97:
Issue #, if available: https://github.com/awslabs/kubernetes-iteration-toolkit/issues/81

#### Liveness Probes:
Adding liveness probes to control plane components. We are defining HTTP liveness probes for these containers:
- apiserver
- controller-manager
- scheduler
- authenticator
- etcd

The kubelet performs a liveness probe for each of these containers at the given interval, and if the server hosting the liveness check returns a success code, the kubelet considers the container to be alive and healthy. If the handler returns a failure code, the kubelet kills the container and restarts it. Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

#### Readiness Probes:
Additionally, we are defining an HTTP readiness probe for the apiserver only. Rather than using the deprecated healthz endpoint, we are using the livez endpoint for the liveness probe and the readyz endpoint for the readiness probe. Readiness probes are only required for pods which accept network traffic through services which is only done for apiserver and etcd pods. Note that we can't add a readiness probe to etcd pods because we rely on the etcd ClusterIP service for initial member discovery:
```
    - --initial-cluster=example-etcd-0=https://example-etcd-0.example-etcd.default.svc.cluster.local:2380,example-etcd-1=https://example-etcd-1.example-etcd.default.svc.cluster.local:2380,example-etcd-2=https://example-etcd-2.example-etcd.default.svc.cluster.local:2380
```
The etcd /health endpoint returns false until the cluster is bootstrapped and has quorum. There would be a circular dependency on the readiness health check passing the endpoints being added behind the etcd ClusterIP service.

#### Testing:
Confirming that liveness probes are correctly configured for each container. Created an example control plane and confirmed no containers were being restarted due to misconfigured or failing liveness probes. Note that we're using hostNetwork set to true for these pods so we are specifying the host as localhost. To test, we exec'ed into one of the apiserver, controller-manager, scheduler, or authenticator pods to hit the health endpoints for each of these components on the node's localhost (alternatively, we could have SSH'ed on the node and hit localhost rather than from inside a pod using the hostNetwork). Similarly to test etcd, we exec'ed into the etcd pod because that pod is running on its own node also using the hostNetwork.

1. apiserver: note that this liveness probe uses an HTTPS liveness probe. Since we're using HTTPS, the kubelet sends an HTTPS request while skipping the certificate verification. The apiserver does not require certs for the /healthz endpoint.
```
Liveness:   http-get https://127.0.0.1:443/livez delay=10s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl -k https://127.0.0.1:443/livez
ok

Readiness:  http-get https://127.0.0.1:443/readyz delay=0s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl -k https://127.0.0.1:443/readyz
ok
```
2. controller-manager: 
```
Liveness:     http-get http://127.0.0.1:10252/healthz delay=10s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl http://127.0.0.1:10252/healthz
ok
```
3. scheduler:
```
Liveness:     http-get http://127.0.0.1:10251/healthz delay=10s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl http://127.0.0.1:10251/healthz
ok
```
4. authenticator:
```
Liveness:       http-get http://127.0.0.1:21363/healthz delay=10s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl http://127.0.0.1:21363/healthz
ok
```
5. etcd:
```
Liveness:       http-get http://127.0.0.1:2381/health delay=10s timeout=5s period=5s #success=1 #failure=5

sh-4.2# curl http://127.0.0.1:2381/health
{"health":"true"}
```


operator/pkg/controllers/etcd/pod.go
```diff
@@ -115,6 +116,19 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
 					},
 				},
 			}},
+			LivenessProbe: &v1.Probe{
+				Handler: v1.Handler{
+					HTTPGet: &v1.HTTPGetAction{
```
ellistarn, line 121: Where do these values come from?

ellistarn: Not sure how useful this is: https://github.com/coreos/etcd-operator/blob/13a1831f9744207b39733bbe91413ca2e389aa40/pkg/util/k8sutil/pod_util.go#L71

operator/pkg/controllers/master/authenticatorconfig.go
```diff
@@ -108,6 +108,20 @@ func (c *Controller) reconcileAuthenticatorDaemonSet(ctx context.Context, contro
 								Name:      "kubeconfig",
 								MountPath: "/var/aws-iam-authenticator/kubeconfig/",
 							}},
+							LivenessProbe: &v1.Probe{
+								Handler: v1.Handler{
+									HTTPGet: &v1.HTTPGetAction{
+										Host:   "127.0.0.1",
+										Scheme: v1.URISchemeHTTP,
+										Path:   "/healthz",
```
ellistarn, line 116: Does this component separate readiness/liveness?

operator/pkg/controllers/master/kubeapiserver.go
```diff
@@ -198,6 +199,19 @@ func apiServerPodSpecFor(controlPlane *v1alpha1.ControlPlane) v1.PodSpec {
 					MountPath: "/var/aws-iam-authenticator/kubeconfig/",
 					ReadOnly:  true,
 				}},
+				LivenessProbe: &v1.Probe{
+					Handler: v1.Handler{
+						HTTPGet: &v1.HTTPGetAction{
+							Host:   "127.0.0.1",
+							Scheme: v1.URISchemeHTTPS,
+							Path:   "/healthz",
```
ellistarn, line 207: likewise, readiness vs liveness: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

operator/pkg/controllers/etcd/pod.go
```diff
@@ -115,6 +116,19 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
...
+						Host:   "127.0.0.1",
+						Scheme: v1.URISchemeHTTP,
+						Path:   "/health",
+						Port:   intstr.FromInt(2381),
+					},
+				},
+				InitialDelaySeconds: 120,
+				TimeoutSeconds:      15,
+				FailureThreshold:    8,
```
prateekgogia, line 130: Similarly, we can reduce TimeoutSeconds and FailureThreshold values so that kubelet can restart them quickly. Also, I checked the default `PeriodSeconds` is 10 seconds maybe we can reduce this as well to be 5 seconds?

operator/pkg/controllers/master/authenticatorconfig.go
prateekgogia: For authenticator we don't connect through a service, so readiness might not be as useful.

operator/pkg/controllers/master/kubeapiserver.go
```diff
@@ -198,6 +199,19 @@ func apiServerPodSpecFor(controlPlane *v1alpha1.ControlPlane) v1.PodSpec {
...
+							Port:   intstr.FromInt(443),
+						},
+					},
+					InitialDelaySeconds: 15,
+					TimeoutSeconds:      15,
+					FailureThreshold:    8,
```
prateekgogia, line 213: Similarly, consider detecting failures earlier and faster.

operator/pkg/controllers/master/authenticatorconfig.go
natherz97: Good point, will add a readiness probes to apiserver and etcd pods since traffic to these pods is sent through services.

operator/pkg/controllers/master/kubeapiserver.go
natherz97: Will move away from healthz endpoint for apiserver and use readyz for readiness and livez for liveness.

operator/pkg/controllers/etcd/pod.go
natherz97: We're also using HTTP probes where we need just need to specify a host and endpoint and need a successful HTTP status code. With command probes, we could use curl or etcdctl but those probes rely on a non-zero exit code so with curl we could have an exit code of zero but get returned a 400 and still need to check the content of what's returned from the endpoint.

operator/pkg/controllers/etcd/pod.go
natherz97: I reduced the InitialDelaySeconds, TimeoutSeconds, FailureThreshold, and Period for all liveness probes.

operator/pkg/controllers/etcd/pod.go
```diff
@@ -115,6 +116,19 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
...
+			LivenessProbe: &v1.Probe{
+				Handler: v1.Handler{
+					HTTPGet: &v1.HTTPGetAction{
+						Host:   "127.0.0.1",
+						Scheme: v1.URISchemeHTTP,
+						Path:   "/health",
+						Port:   intstr.FromInt(2381),
+					},
+				},
+				InitialDelaySeconds: 120,
```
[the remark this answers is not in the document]
natherz97: Yes, the cluster needs to have quorum for the etcd health check to succeed. If one etcd pod is terminated, the health check will continue to succeed on the other two pods.

operator/pkg/controllers/etcd/pod.go
```diff
@@ -115,6 +116,20 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
 					},
 				},
 			}},
+			LivenessProbe: &v1.Probe{
+				Handler: v1.Handler{
+					HTTPGet: &v1.HTTPGetAction{
+						Host:   "127.0.0.1",
+						Scheme: v1.URISchemeHTTP,
+						Path:   "/health",
+						Port:   intstr.FromInt(2381),
```
prateekgogia, line 125: Thoughts on just passing the string directly to Port
```
Port: "2381",
```

prateekgogia: nm, checked offline with Nathan Port field is `intstr.IntOrString`

prateekgogia (approved):
LGTM!

ellistarn (approved):

Issue #12, opened 2021-07-20
[operator] etcd cluster should be resilient to node/pod failures


prateekgogia:
Creating this issue for tracking purposes.

There are two ways to configure the etcd pod- 
- When etcd data directory `/var/lib/etcd` node volume is mounted to the etcd pod, following issues are seen -
    - When an etcd cluster is created and quorum is established, If a node gets terminated new node will be created by karpenter and will have a different peer/member ID and quorum is not established.
    - When an etcd cluster is created and quorum is established, and if statefulset is deleted and recreated on the same set of nodes, new etcd pods come up with different IDs and quorum is never established.

- When etcd data directory `/var/lib/etcd` is not mounted to the etcd pod
   - If a pod gets restarted, quorum is not established and we see the following error -
      `member 86adba6a3f2b2193 has already been bootstrapped`

prateekgogia:
Ref: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/testing-manifests/statefulset/etcd/statefulset.yaml

PR #132, opened 2022-02-18
Make etcd resilient to pod, node, and volume failures

natherz97:
Issue #, if available: https://github.com/awslabs/kubernetes-iteration-toolkit/issues/12

#### Description: 
Make etcd resilient to pod, node, and volume failures. On initial etcd startup, we set initial-cluster-state to new and start the three etcd members. On a member restart (such as when a pod is recreated), the initial member will rejoin using initial-cluster-state existing. In this case, we retain the etcd data-dir which is a HostPath volume on the existing node which allows the member to rejoin. On a permanent member failure (such as when an etcd node is deleted), the old member will have to be removed and a new member will have to be added. In this case, we lose the contents of the etcd data-dir since it is a HostPath volume on the deleted node.

When we move to EBS PVs rather than HostPath volumes for the etcd data-dir, the pod restart and node deletion cases will both allow the existing member to rejoin the cluster since in both cases the contents of the data-dir for that member would be retained. Additionally, when we have EBS PVs, we won't need to update the advertise peer URL when the pod starts on a new node because we're using the pod DNS name which will always be the same for a given member. When we move to EBS PVs, the only time we'll need to add/remove members is when the backing EBS volume for a member is deleted or corrupted.

#### Testing:
New cluster initial start-up:
```
etcd_bootstrap: sleeping for 5 seconds
etcd_bootstrap: using etcdctl endpoints --endpoints https://example-etcd-0.example-etcd.default.svc.cluster.local:2379,https://example-etcd-1.example-etcd.default.svc.cluster.local:2379,https://example-etcd-2.example-etcd.default.svc.cluster.local:2379
{"level":"warn","ts":"2022-02-18T23:01:17.572Z","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-c33722da-b922-402d-9c9e-2f35e02e5a31/example-etcd-0.example-etcd.default.svc.cluster.local:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp: lookup example-etcd-1.example-etcd.default.svc.cluster.local on 10.100.0.10:53: no such host\""}
Error: context deadline exceeded
etcd_bootstrap: cluster state new
etcd_bootstrap: member info
etcd_bootstrap: current node example-etcd-1
ls: cannot access /var/lib/etcd/member: No such file or directory
etcd_bootstrap: contents of /var/lib/etcd/member
```
Deleting an etcd pod which will restart an existing member:
```
etcd_bootstrap: sleeping for 5 seconds
etcd_bootstrap: using etcdctl endpoints --endpoints https://example-etcd-0.example-etcd.default.svc.cluster.local:2379,https://example-etcd-1.example-etcd.default.svc.cluster.local:2379,https://example-etcd-2.example-etcd.default.svc.cluster.local:2379
etcd_bootstrap: cluster state existing
etcd_bootstrap: member info 85c52305520b53a9, started, example-etcd-2, https://example-etcd-2.example-etcd.default.svc.cluster.local:2380, https://example-etcd-2.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
d2ea75be6ce30272, started, example-etcd-0, https://example-etcd-0.example-etcd.default.svc.cluster.local:2380, https://example-etcd-0.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
f9fb2447635a312f, started, example-etcd-1, https://example-etcd-1.example-etcd.default.svc.cluster.local:2380, https://example-etcd-1.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
etcd_bootstrap: current node example-etcd-0
etcd_bootstrap: contents of /var/lib/etcd/member snap
wal
```
Deleting an etcd node which will make us lose the HostPath data-dir and require removing and adding a new member:
```
etcd_bootstrap: sleeping for 5 seconds
etcd_bootstrap: using etcdctl endpoints --endpoints https://example-etcd-0.example-etcd.default.svc.cluster.local:2379,https://example-etcd-1.example-etcd.default.svc.cluster.local:2379,https://example-etcd-2.example-etcd.default.svc.cluster.local:2379
etcd_bootstrap: cluster state existing
etcd_bootstrap: member info 85c52305520b53a9, started, example-etcd-2, https://example-etcd-2.example-etcd.default.svc.cluster.local:2380, https://example-etcd-2.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
d2ea75be6ce30272, started, example-etcd-0, https://example-etcd-0.example-etcd.default.svc.cluster.local:2380, https://example-etcd-0.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
f9fb2447635a312f, started, example-etcd-1, https://example-etcd-1.example-etcd.default.svc.cluster.local:2380, https://example-etcd-1.example-etcd.default.svc.cluster.local:2379,https://example-etcd.default.svc.cluster.local:2379, false
etcd_bootstrap: current node example-etcd-0
ls: cannot access /var/lib/etcd/member: No such file or directory
etcd_bootstrap: contents of /var/lib/etcd/member
ls: cannot access /var/lib/etcd/member: No such file or directory
etcd_bootstrap: we're removing the current member and adding back
{"level":"warn","ts":"2022-02-18T23:08:53.362Z","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"endpoint://client-641e1f99-bd82-4106-8079-12e6724920d1/example-etcd-0.example-etcd.default.svc.cluster.local:2379","attempt":0,"error":"rpc error: code = DeadlineExceeded desc = latest balancer error: all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 192.168.147.11:2379: connect: connection refused\""}
https://example-etcd-2.example-etcd.default.svc.cluster.local:2379 is healthy: successfully committed proposal: took = 10.8448ms
https://example-etcd-1.example-etcd.default.svc.cluster.local:2379 is healthy: successfully committed proposal: took = 14.178022ms
https://example-etcd-0.example-etcd.default.svc.cluster.local:2379 is unhealthy: failed to commit proposal: context deadline exceeded
Error: unhealthy cluster
etcd_bootstrap: removing member d2ea75be6ce30272
Member d2ea75be6ce30272 removed from cluster b8de25a771469c4c
etcd_bootstrap: adding member back
Member 65b3d4c2252a006e added to cluster b8de25a771469c4c

ETCD_NAME="example-etcd-0"
ETCD_INITIAL_CLUSTER="example-etcd-0=https://example-etcd-0.example-etcd.default.svc.cluster.local:2380,example-etcd-2=https://example-etcd-2.example-etcd.default.svc.cluster.local:2380,example-etcd-1=https://example-etcd-1.example-etcd.default.svc.cluster.local:2380"
ETCD_INITIAL_ADVERTISE_PEER_URLS="https://example-etcd-0.example-etcd.default.svc.cluster.local:2380"
ETCD_INITIAL_CLUSTER_STATE="existing"
```


operator/pkg/controllers/etcd/pod.go
```diff
@@ -222,3 +217,56 @@ func nodeSelector(clusterName string) map[string]string {
 	return functional.UnionStringMaps(labelsFor(clusterName),
 		map[string]string{object.ControlPlaneLabelKey: clusterName, instanceTypeLabelKey: instanceTypeLabelDefaultValue})
 }
+
+var etcdBootstrapCmd = `
+echo "etcd_bootstrap: sleeping for 5 seconds"
+sleep 5
```
ellistarn, line 223: What are you racing against here?

operator/pkg/controllers/etcd/pod.go
```diff
@@ -187,10 +169,23 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
...
+
+func etcdctlEndpoints(controlPlane *v1alpha1.ControlPlane) string {
+	return endPointFlagGenerator(controlPlane, "https://%[1]s-etcd-%[2]d.%[1]s-etcd.%[3]s.svc.cluster.local:2379")
+}
+
+func endPointFlagGenerator(controlPlane *v1alpha1.ControlPlane, format string) string {
 	nodes := make([]string, 0)
 	for i := 0; i < controlPlane.Spec.Etcd.Replicas; i++ {
-		nodes = append(nodes, fmt.Sprintf("%[1]s-etcd-%[2]d=https://%[1]s-etcd-%[2]d.%[1]s-etcd.%[3]s.svc.cluster.local:2380", controlPlane.ClusterName(), i, controlPlane.Namespace))
+		nodes = append(nodes, fmt.Sprintf(format, controlPlane.ClusterName(), i, controlPlane.Namespace))
```
ellistarn, line 188: Usually we pass in order of (namespace, name), since resources are denoted this way in k8's paths (e.g. parent/child in rest).

operator/pkg/controllers/etcd/pod.go
```diff
@@ -187,10 +169,23 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
 	}
 }
 
+func generateEtcdBootstrapCmd(controlPlane *v1alpha1.ControlPlane) string {
```
ellistarn, line 172: Some naming nits:
- I'd drop the word `generate` and just noun based getters (e.g. foo.generateBar() -> foo.bar())
- You're stuttering with the word `etcd` (e.g. `etcd.generateEtcd...`). We're already contextualized to etcd in this package. 

I wonder if we should be attaching this behavior to the controller, rather than unowned static methods.
`controller.bootstrapCommand(controlPlane *v1alpha1.ControlPlane)`

I can see leaving this as a static, as long as this package remains tightly scoped and we don't leak these functions anywhere.

operator/pkg/controllers/etcd/pod.go
```diff
@@ -187,10 +169,23 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
...
+	return fmt.Sprintf(etcdBootstrapCmd, etcdctlEndpoints(controlPlane), advertizePeerURL(controlPlane),
+		initialClusterFlag(controlPlane), advertizeClusterURL(controlPlane))
+}
+
 func initialClusterFlag(controlPlane *v1alpha1.ControlPlane) string {
+	return endPointFlagGenerator(controlPlane, "%[1]s-etcd-%[2]d=https://%[1]s-etcd-%[2]d.%[1]s-etcd.%[3]s.svc.cluster.local:2380")
+}
+
+func etcdctlEndpoints(controlPlane *v1alpha1.ControlPlane) string {
```
ellistarn, line 181: similar, consider using `endpoints`.

operator/pkg/controllers/etcd/pod.go
```diff
@@ -187,10 +169,23 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
...
+	return endPointFlagGenerator(controlPlane, "https://%[1]s-etcd-%[2]d.%[1]s-etcd.%[3]s.svc.cluster.local:2379")
+}
+
+func endPointFlagGenerator(controlPlane *v1alpha1.ControlPlane, format string) string {
 	nodes := make([]string, 0)
 	for i := 0; i < controlPlane.Spec.Etcd.Replicas; i++ {
```
ellistarn, line 187: I'm not sure abstracting this for loop helps readability. It took me a while to figure out how this was wired up. I think better naming could help (e.g. [curry](https://en.wikipedia.org/wiki/Currying), but it may just be simpler to unroll the for loop and repeat yourself, since it's only used in two locations.

prateekgogia (changes requested):
Great work with this one!

natherz97:
Responding to comments that are marked as outdated due to force pushing:
- I read this isn't necessary after etcd3.4, what version are we on?
You're right, setting ETCDCTL_API=3 is not needed since we're using v3.4.16.
- I dropped the word generate and etcd from function naming. Other components like apiserver, KSH, etc. are following this pattern with using static functions so I'll keep these functions static for consistency.
- For etcdctl endpoints, we switched to using the service DNS name rather than individual endpoints so there's no longer a need to abstract out the endPointFlagGenerator function.
- High level question -- should we just invest in ebs volumes for etcd?
I updated the overview to be consistent with EBS volumes. Now with EBS volumes, etcd members can rejoin the cluster after pod or node deletions (previously the host path volume would be deleted on node deletion). The only case where a member has to be removed and restarted is when the underlying EBS volume is deleted.
- What are you racing against here?
I removed the sleep time in the latest revision after re-executing the test cases with EBS volumes and it was not needed for members to successfully call etcdctl.
- Usually we pass in order of (namespace, name), since resources are denoted this way in k8's paths (e.g. parent/child in rest).
My commit no longer touches this line so I kept the current ordering. I think keeping the ordering consistent with Sprintf arguments is better for readability than following namespace, name ordering.

ellistarn (dismissed):

operator/pkg/utils/patch/patch.go
```diff
@@ -23,6 +23,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
+var blockListedAdditionalContainerFlags = map[string]map[string]bool{
```
prateekgogia, line 26: nit: naming `blockListedAdditionalContainerFlags` -> `blockListedFlags`

operator/pkg/controllers/etcd/pod.go
```diff
@@ -80,13 +80,15 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
 			}, {
 				Name:      "etcd-server-certs",
 				MountPath: "/etc/kubernetes/pki/etcd/server",
+			}, {
+				Name:      "etcd-bootstrap",
+				MountPath: "/etc/kubernetes/",
 			}},
-			Command: []string{"etcd"},
 			Args: []string{
+				"./etc/kubernetes/bootstrap.sh",
```
prateekgogia, line 88: Have we tried have this script as part of the `Command` field?

operator/pkg/controllers/etcd/bootstrapconfig.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func bootstrapConfigMapName(clusterName string) string {
+	return fmt.Sprintf("%s-etcd-bootstrap", clusterName)
+}
+
+// bootstrapAdvertizePeerURL is the same as advertizePeerURL but uses curly brackets rather than parentheses around
+// the NODE_ID environment variable. When executing the bootstrap Configmap as a bash script in the etcd container,
+// we will get this error if we try using parentheses: ./etc/kubernetes/bootstrap.sh: line 27: NODE_ID: command not found
+func bootstrapAdvertizePeerURL(controlPlane *v1alpha1.ControlPlane) string {
```
prateekgogia, line 43: Does this need to be a function, can just use a variable like we are doing for `etcdctlEndpoint`?

prateekgogia (approved):
Great work on this one!

PR #134, opened 2022-02-25
support for adding additional operators elb, csi, karpenter, kit

prateekgogia:
Issue #, if available:

Description of changes:
- Add DNS addon to substrate cluster
- Switched API server port for substrate node to `8443`, we reuse the same subnets, security groups for tenant clusters. Substrate cluster port 8443 is accessible to public. However, 443 is not open. It can only be accessed through an NLB
- Added all the IAM permissions needed by controllers running on substrate node.


substrate/pkg/controller/substrate/cluster/addons/dns.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+
+	"github.com/awslabs/kit/substrate/pkg/apis/v1alpha1"
+	"github.com/awslabs/kit/substrate/pkg/controller/substrate/cluster"
+	"k8s.io/kubernetes/cmd/kubeadm/app/phases/addons/dns"
+	"k8s.io/kubernetes/cmd/kubeadm/app/util/kubeconfig"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type DNS struct {
```
ellistarn, line 28: I'm a little curious if its worth colocating all of these addons into the same controller `addons.go`, since it's a bunch of interdependent calls to kubeadm. It may be marginally faster this way, but it appears as if there is more complexity than there is.

substrate/pkg/controller/substrate/infrastructure/securitygroup.go
```diff
@@ -44,9 +44,14 @@ func (s *SecurityGroup) Create(ctx context.Context, substrate *v1alpha1.Substrat
 		GroupId: securityGroup.GroupId,
 		IpPermissions: []*ec2.IpPermission{{
 			IpProtocol: aws.String("tcp"),
-			FromPort:   aws.Int64(443),
-			ToPort:     aws.Int64(443),
+			FromPort:   aws.Int64(8443),
+			ToPort:     aws.Int64(8443),
 			IpRanges:   []*ec2.IpRange{{CidrIp: aws.String("0.0.0.0/0")}},
+		}, {
+			IpProtocol:       aws.String("-1"),
```
ellistarn, line 51: What's going on here? Can you add a comment?

substrate/pkg/controller/substrate/infrastructure/subnets.go
```diff
@@ -97,12 +97,22 @@ func (s *Subnets) ensureSubnet(ctx context.Context, substrate *v1alpha1.Substrat
 	if len(describeSubnetsOutput.Subnets) > 0 {
 		logging.FromContext(ctx).Infof("Found subnet %s", aws.StringValue(name))
 		return describeSubnetsOutput.Subnets[0], nil
+
+	}
+	// these tags are required by ELB controller to discover these subnets to configure ELB
+	tags := []*ec2.Tag{
+		{Key: aws.String("kubernetes.io/cluster/" + substrate.Name), Value: aws.String("shared")},
```
ellistarn, line 104: IIUC this is no longer necessary.

substrate/pkg/controller/substrate/cluster/instanceprofile.go
```diff
@@ -186,6 +185,91 @@ func desiredRolesFor(substrate *v1alpha1.Substrate) []role {
...
+						"iam:GetRole",
+						"iam:GetInstanceProfile",
+						"iam:CreateServiceLinkedRole",
+						"iam:ListServerCertificates",
+						"iam:GetServerCertificate",
+						"shield:GetSubscriptionState",
+						"shield:DescribeProtection",
+						"shield:CreateProtection",
+						"shield:DeleteProtection",
+						"waf-regional:GetWebACL",
```
ellistarn, line 262: Where did all these come from?

prateekgogia: These are the permissions required by ELB controller and CSI driver, I got these from their policy files.

substrate/pkg/controller/substrate/infrastructure/subnets.go
prateekgogia: Yeah Thanks, I missed the last line `The cluster tag is not required in v2.1.2 and newer releases.` here -https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/deploy/subnet_discovery/

substrate/pkg/controller/substrate/cluster/addons/dns.go
prateekgogia: Yeah we can explore but not all addons are straightforward as this on. The new functionality to deploy helm chart as addons I just added are little more complex.

substrate/pkg/controller/substrate/cluster/addons/operators.go
```diff
@@ -0,0 +1,118 @@
...
+			"aws": map[string]interface{}{"defaultInstanceProfile": discovery.Name(substrate, cluster.TenantControlPlaneNodeRole)}}},
+		{awsEBSCSIDriverChart, "kube-system", "aws-ebs-csi-driver", map[string]interface{}{"controller": map[string]interface{}{"replicaCount": "1"}}},
+		{awsLBControllerChart, "kube-system", "aws-load-balancer-controller", map[string]interface{}{"clusterName": substrate.Name, "replicaCount": "1"}},
+	}
+	errs := make([]error, len(charts))
+	workqueue.ParallelizeUntil(ctx, len(charts), len(charts), func(i int) {
+		errs[i] = h.ensure(charts[i], substrate)
+	})
+	if err := multierr.Combine(errs...); err != nil {
+		return reconcile.Result{Requeue: true}, err
```
ellistarn, line 70: This abuses the semantic of reconcile.Result. The go idiom of "result or error" expects an empty first parameter if error is defined. All consuming code should ignore the value of your reconcile.Result if error is defined.

substrate/pkg/utils/discovery/tags.go
```diff
@@ -26,13 +26,13 @@ const (
 	OwnerTagKey = "kit.aws/substrate"
 )
 
-func Tags(substrate *v1alpha1.Substrate, resource string, name *string) []*ec2.TagSpecification {
+func Tags(substrate *v1alpha1.Substrate, resource string, name *string, additionalTags ...*ec2.Tag) []*ec2.TagSpecification {
```
ellistarn, line 29: I think we should do the append outside, instead of in this method as a variadic.

substrate/pkg/controller/substrate/infrastructure/subnets.go
```diff
@@ -97,12 +97,21 @@ func (s *Subnets) ensureSubnet(ctx context.Context, substrate *v1alpha1.Substrat
...
+	if subnetSpec.Public {
+		// these tags are required by Karpenter controller to discover these subnets to provision new node
+		tags = append(tags, &ec2.Tag{Key: aws.String("karpenter.sh/discovery"), Value: aws.String(substrate.Name)})
 	}
 	createSubnetsOutput, err := s.EC2.CreateSubnetWithContext(ctx, &ec2.CreateSubnetInput{
 		AvailabilityZone:  aws.String(subnetSpec.Zone),
 		CidrBlock:         aws.String(subnetSpec.CIDR),
 		VpcId:             substrate.Status.Infrastructure.VPCID,
-		TagSpecifications: discovery.Tags(substrate, ec2.ResourceTypeSubnet, name),
+		TagSpecifications: discovery.Tags(substrate, ec2.ResourceTypeSubnet, name, tags...),
```
ellistarn, line 114: I think it's more idiomatic to do
```
append(discovery.Tags(...), tags...)
```

substrate/pkg/utils/discovery/tags.go
prateekgogia: here the `Tags` function returns `[]*ec2.TagSpecification` and we need to append `&ec2.Tag{}` in tagSpecification slice and at the caller side in subnets where we need this functionality it will look like this-

```
tags := []*ec2.Tag{a:b}
if true{
   tags = append(tags, []*ec2.Tag{c:d}
}
t := discovery.Tags(substrate, ec2.ResourceTypeSubnet, name)
t[0].Tags = append(t[0].Tags, tags)
```
which isn't very clean

substrate/pkg/controller/substrate/infrastructure/subnets.go
prateekgogia: Answered above where the function is defined.

ellistarn (approved):

PR #140, opened 2022-03-08
Add debug flag for CLI

prateekgogia:
Issue #, if available:

Description of changes:
- Adds a debug flag for the cli
- Fix the cli binary name to be kitctl
- Convert some info logs to debug logs specifically once related to discovery

### DIFF
--- a/substrate/README.md
+++ b/substrate/README.md
@@ -1,39 +1,106 @@
-# KITCLI
+# kitctl
 
-## Installing KITCLI
+## Installing kitctl
 
 ```bash
 brew tap awslabs/kit https://github.com/awslabs/kubernetes-iteration-toolkit.git
-brew install kitcli
+brew install kitctl
 ```
 
 ## Usage
+kitctl helps provision an AWS infrastructure environment to deploy Kubernetes clusters using kit-operator. It runs a single node kubernetes cluster in a VPC installed with all the required controllers like Karpeneter, KIT-operator, ELB controller and EBS CSI Driver to manage the lifecycle of clusters. kitctl also creates the necessary IAM permissions required for these controllers.
+To get started make sure you have admin access to AWS.
 
-```sh
-kitcli apply
-kitcli delete
+### Bootstrap an environment in AWS to run Kubernetes clusters using kit-operator
+
+```bash
+kitctl bootstrap
+```
+> Set KUBECONFIG to access the environment with the kubeconfig location provided from this command
+
+### Configure Karpenter to provision nodes for the Kubernetes cluster control plane
+
+```bash
+GUEST_CLUSTER_NAME=foo # Desired Cluster name
+SUBSTRATE_CLUSTER_NAME=test-substrate
 ```
 
-```shell
-cat <<EOF | kitcli apply -f -
-apiVersion: kit.sh/v1alpha1
-kind: Substrate
+```bash
+cat <<EOF | kubectl apply -f -
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
 metadata:
-  name: ${USER}
+  name: master-nodes
 spec:
-  instanceType: c5.2xlarge
----
-apiVersion: kit.sh/v1alpha1
-kind: Cluster
+  kubeletConfiguration:
+    clusterDNS:
+      - "10.96.0.10"
+  labels:
+    kit.k8s.sh/app: ${GUEST_CLUSTER_NAME}-apiserver
+    kit.k8s.sh/control-plane-name: ${GUEST_CLUSTER_NAME}
+  provider:
+    instanceProfile: kit-${SUBSTRATE_CLUSTER_NAME}-tenant-controlplane-node-role
+    subnetSelector:
+      karpenter.sh/discovery: ${SUBSTRATE_CLUSTER_NAME}
+    securityGroupSelector:
+      kubernetes.io/cluster/${SUBSTRATE_CLUSTER_NAME}: owned
+    tags:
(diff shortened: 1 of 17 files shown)


substrate/README.md
```diff
@@ -1,39 +1,106 @@
...
-apiVersion: kit.sh/v1alpha1
-kind: Substrate
+### Configure Karpenter to provision nodes for the Kubernetes cluster control plane
+
+```bash
+GUEST_CLUSTER_NAME=foo # Desired Cluster name
+SUBSTRATE_CLUSTER_NAME=test-substrate
+```
+
+```bash
```
ellistarn, line 28: Let's move this into the bootstrap call soon.

substrate/cmd/bootstrap.go
```diff
@@ -25,18 +26,30 @@ import (
 	"knative.dev/pkg/logging"
 )
 
-func init() {
-	rootCmd.AddCommand(&cobra.Command{
-		Use:   "apply",
-		Short: "Apply an environment for testing. Will reconnect if the environment already exists.",
+func bootstrapCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "bootstrap",
```
ellistarn, line 31: Should we just use the terms create/delete? Since we're creating/deleting an env? Or do we imagine many other commands in kitcli? My assumption is that you should interact with guest clusters using kubectl.

substrate/cmd/bootstrap.go
```diff
@@ -56,5 +69,5 @@ func Apply(cmd *cobra.Command, args []string) {
 		logging.FromContext(ctx).Error(err.Error())
 		return
 	}
-	logging.FromContext(ctx).Infof("Applied substrate %s after %s", name, time.Since(start))
+	logging.FromContext(ctx).Infof("✅ Bootstrap finished, substrate name %s time taken %s", name, time.Since(start))
```
ellistarn, line 72: nit: this log line doesn't as plain english, especially "name %s time taken %s". I could see something like "Bootstrapped %s after %d".

substrate/cmd/delete.go
```diff
@@ -24,18 +24,16 @@ import (
 	"knative.dev/pkg/logging"
 )
 
-func init() {
-	rootCmd.LocalFlags().StringVar(&options.File, "name", "", "Name for the environment")
-	rootCmd.LocalFlags().StringVarP(&options.File, "file", "f", "", "Configuration file for the environment")
-	rootCmd.AddCommand(&cobra.Command{
+func deleteCommand() *cobra.Command {
```
ellistarn, line 27: I like calling this delete. Makes me think the other command should be "create" instead of "bootstrap"

substrate/pkg/controller/substrate/cluster/config.go
```diff
@@ -103,8 +103,9 @@ func (c *Config) Create(ctx context.Context, substrate *v1alpha1.Substrate) (rec
 		aws.StringValue(discovery.Name(substrate)), path.Join(ClusterCertsBasePath, aws.StringValue(discovery.Name(substrate))))); err != nil {
 		return reconcile.Result{}, fmt.Errorf("uploading to S3 %w", err)
 	}
-	logging.FromContext(ctx).Infof("Uploaded cluster configuration to s3://%s", aws.StringValue(discovery.Name(substrate)))
+	logging.FromContext(ctx).Debugf("Uploaded cluster configuration to s3://%s", aws.StringValue(discovery.Name(substrate)))
 	substrate.Status.Cluster.KubeConfig = ptr.String(path.Join(ClusterCertsBasePath, aws.StringValue(discovery.Name(substrate)), kubeconfigFile))
+	logging.FromContext(ctx).Infof("To access substrate cluster, export KUBECONFIG=%s", aws.StringValue(substrate.Status.Cluster.KubeConfig))
```
ellistarn, line 108: Can you add a new line such that the full command is easily selectable?

substrate/cmd/delete.go
```diff
@@ -45,5 +43,5 @@ func Delete(cmd *cobra.Command, args []string) {
 		logging.FromContext(ctx).Error(err.Error())
 		return
 	}
-	logging.FromContext(ctx).Infof("Deleted substrate %s after %s", name, time.Since(start))
+	logging.FromContext(ctx).Debugf("Deleted substrate %s after %s", name, time.Since(start))
```
ellistarn, line 46: Should be info, I think.

substrate/cmd/bootstrap.go
prateekgogia: I plan on adding a `create` command for cluster creation, that would generate a control-plane spec. Something like `kitctl create cluster foo`

substrate/cmd/delete.go
prateekgogia: `delete` conflicts with the delete function call made as part of this command

ellistarn: Oh I'm not suggesting calling it delete(). I'm just saying that I like your naming of `kitctl delete`

substrate/cmd/bootstrap.go
ellistarn: This will have strange parity with delete, since create and delete aren't inverted from each other.

ellistarn (approved):

substrate/cmd/bootstrap.go
prateekgogia: Yeah I noticed that as well, this is what I was thinking the flow will look like, wdyt?
```
kictl bootstrap
kictl create cluster foo
kictl delete cluster foo
kitctl cleanup
```

PR #150, opened 2022-03-10
Adds support for KMS secret encryption in operator

prateekgogia:
Issue #, if available:

Description of changes:


operator/pkg/apis/controlplane/v1alpha1/controlplane.go
```diff
@@ -56,6 +56,7 @@ type Etcd struct {
 // custom flags for components running on master nodes like apiserver, KCM and
 // scheduler.
 type MasterSpec struct {
+	KMSKeyARN         *string    `json:"kmskey,omitempty"`
```
ellistarn, line 59: Should we use keyname instead of arn? Will work across users/accounts.

operator/pkg/controllers/master/master.go
```diff
@@ -60,6 +60,8 @@ func (c *Controller) Reconcile(ctx context.Context, controlPlane *v1alpha1.Contr
 		c.reconcileScheduler,
 		c.reconcileAuthenticator,
 		c.iamController.Reconcile,
+		c.reconcileEncryptionProviderConfig,
```
ellistarn, line 63: I'd like to (eventually) migrate to making these objects instead of methods. Let's punt for now.
```suggestion
		c.encryptionProvider,
```

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
prateekgogia: This is the `arn` we need to pass to aws-encryption-provider which gets supplied [here](https://github.com/kubernetes-sigs/aws-encryption-provider/blob/dee8cbcb643284037574f32181ed320e5ced0276/vendor/github.com/aws/aws-sdk-go/service/kms/api.go#L9823), is the keyname same as alias?

operator/pkg/controllers/master/master.go
prateekgogia: We have an `reconcileEncryptionProviderConfig` and `reconcileEncryptionProvider`, won't just having `encryptionProvider` be confusing which objects/resource is reconciler

substrate/cmd/root.go
```diff
@@ -66,13 +66,15 @@ func newRootCmd(args []string) *cobra.Command {
 var options = &Options{}
 
 type Options struct {
-	file  string
-	debug bool
-	help  bool
+	file          string
+	debug         bool
+	substrateName string
```
ellistarn, line 71:
```suggestion
	name string
```

substrate/README.md
```diff
@@ -14,15 +14,16 @@ To get started make sure you have admin access to AWS.
 ### Bootstrap an environment in AWS to run Kubernetes clusters using kit-operator
 
 ```bash
-kitctl bootstrap
+export AWS_REGION=us-west-2
+kitctl bootstrap -n substrate-$(whoami) # Environment name
```
ellistarn, line 18: Let's avoid using the word substrate where possible. 
```suggestion
kitctl bootstrap -n kitctl-$(whoami) # Environment name
```

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
ellistarn: >	// To specify a CMK, use its key ID, Amazon Resource Name (ARN), alias name,
	// or alias ARN. When using an alias name, prefix it with "alias/". To specify
	// a CMK in a different AWS account, you must use the key ARN or alias ARN.

Looks like all of them work. Perhaps we can call this `KeyId` for parity?

substrate/README.md
```diff
@@ -105,7 +106,7 @@ kubectl delete controlplane ${GUEST_CLUSTER_NAME}
 - To clean up the AWS environment
 
 ```bash
-kitctl helps provision an AWS infrastructure environment to deploy Kubernetes clusters using kit-operator. It runs a single node kubernetes cluster with all the required controllers like Karpeneter, KIT-operator, ELB controller and EBS CSI Driver to manage the lifecycle of clusters. kitctl also creates the necessary delete
+kitctl delete <kitctl-$(whoami)> # Optional environment name
```
ellistarn, line 109:
```suggestion
kitctl delete kitctl-$(whoami) # Optional environment name
```

substrate/cmd/kitctl/delete.go
```diff
@@ -39,9 +36,8 @@ func deleteCommand() *cobra.Command {
 func delete(cmd *cobra.Command, args []string) {
 	ctx := cmd.Context()
 	start := time.Now()
-	u, err := user.Current()
-	runtime.Must(err)
-	name := fmt.Sprintf("kitctl-%s", u.Username)
+	name := parseName(ctx, args)
+	logging.FromContext(ctx).Infof("Starting cleanup of %q", name)
```
ellistarn, line 40:
```suggestion
	logging.FromContext(ctx).Infof("Deleting %q", name)
```

substrate/cmd/kitctl/bootstrap.go
```diff
@@ -52,9 +49,8 @@ func bootstrap(cmd *cobra.Command, args []string) {
 	}
 	ctx := cmd.Context()
 	start := time.Now()
-	u, err := user.Current()
-	runtime.Must(err)
-	name := fmt.Sprintf("kitctl-%s", u.Username)
+	name := parseName(ctx, args)
+	logging.FromContext(ctx).Infof("Starting bootstrap of %q", name)
```
ellistarn, line 53:
```suggestion
	logging.FromContext(ctx).Infof("Bootstrapping %q", name)
```

substrate/cmd/kitctl/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"fmt"
+	"os/user"
+
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"knative.dev/pkg/logging"
+)
+
+func parseName(ctx context.Context, args []string) string {
+	if len(args) > 1 {
```
ellistarn, line 27: There's gotta be a more idiomatic way to do this with cobra.

substrate/pkg/controller/substrate/controller.go
```diff
@@ -100,6 +100,7 @@ func (c *Controller) Reconcile(ctx context.Context, substrate *v1alpha1.Substrat
 			}
 			result, err := f(ctx, mutable)
 			if err != nil {
+				logging.FromContext(ctx).Debugf("Reconciling %s, error %v", reflect.ValueOf(resource).Elem().Type(), err)
```
ellistarn, line 103: Why do we need to log error here? If this isbubbled up at the end w/ the multierr.Combine, do we log the same error twice?

prateekgogia: In some cases, we just get stuck in a for loop without exiting. This is to catch such scenarios. Recent scenario we ran into was when AWS_REGION is not set. Now that we have this `for ctx.Err() == nil`, don't need this log anymore. I will remove it.

operator/pkg/apis/controlplane/v1alpha1/controlplane.go
```diff
@@ -56,6 +56,7 @@ type Etcd struct {
 // custom flags for components running on master nodes like apiserver, KCM and
 // scheduler.
 type MasterSpec struct {
+	KeyId             *string    `json:"keyId,omitempty"`
```
ellistarn, line 59: Sorry to keep picking on this, but given the context, I think we probably need to specify KMSKeyId. It's currently ambiguous e.g. SSHKeyId

ellistarn (approved):

PR #153, opened 2022-03-16
Upgrade Karpenter charts to v0.7.1

prateekgogia:
Issue #, if available:

Description of changes:

### DIFF
--- a/substrate/README.md
+++ b/substrate/README.md
@@ -21,66 +21,10 @@ kitctl bootstrap kitctl-$(whoami) # Optional environment name
 ```
 > Set KUBECONFIG to access the environment with the kubeconfig location provided from this command
 
-### Configure Karpenter to provision nodes for the Kubernetes cluster control plane
-
-```bash
-GUEST_CLUSTER_NAME=foo # Desired Cluster name
-SUBSTRATE_CLUSTER_NAME=kitctl-$(whoami)
-```
-
-```bash
-cat <<EOF | kubectl apply -f -
-apiVersion: karpenter.sh/v1alpha5
-kind: Provisioner
-metadata:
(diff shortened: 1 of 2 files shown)


substrate/pkg/controller/substrate/cluster/addons/karpenter.go
```diff
@@ -39,8 +39,17 @@ kind: Provisioner
 metadata:
   name: default
 spec:
+  requirements:
+  - key: kit.k8s.sh/app
```
ellistarn, line 43: Why do we need both?

ellistarn (approved):

substrate/README.md
```diff
@@ -21,66 +21,10 @@ kitctl bootstrap kitctl-$(whoami) # Optional environment name
...
-    tags:
-      kit.aws/substrate: ${SUBSTRATE_CLUSTER_NAME}
-  ttlSecondsAfterEmpty: 30
-EOF
-```
-
 ### Provision a Kubernetes cluster control plane
 
 ```bash
+GUEST_CLUSTER_NAME=foo # Desired Cluster name
```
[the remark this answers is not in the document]
prateekgogia: I just wanted to highlight this can be anything a user wants, in case they create multiple guest clusters.

substrate/pkg/controller/substrate/cluster/addons/karpenter.go
prateekgogia: Currently we have two nodeselectors on the API server and etcd pods to one to identify which control plane they belong to and another one to distinguish among themselves. This we will need to revisit at some point as these node selectors will change.

PR #160, opened 2022-03-23
Add support for mixed instances for etcd clusters

prateekgogia:
Issue #, if available:

Description of changes:

### DIFF
--- a/operator/docs/examples/etcd-with-mixed-instances.yaml
+++ b/operator/docs/examples/etcd-with-mixed-instances.yaml
@@ -1,0 +1,50 @@
+# Steps to create a cluster with etcd running on mixed instances
+
+## Create hollow pods to pre-provision mixed instances for etcd pods
+```bash
+ZONES=("a" "b" "c")
+SIZES=("m5.large" "t3.medium" "c5.2xlarge")
+CONTROL_PLANE=foo 
+for i in {1..3}; do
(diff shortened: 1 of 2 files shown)


operator/docs/examples/etcd-with-mixed-instances.yaml
```diff
@@ -0,0 +1,50 @@
...
+```
+
+## Provision kubernetes cluster
+```bash
+CONTROL_PLANE=foo
+cat <<EOF | kubectl apply -f -
+apiVersion: kit.k8s.sh/v1alpha1
+kind: ControlPlane
+metadata:
+  name: $CONTROL_PLANE # Desired Cluster Name
```
ellistarn, line 36: I'd just inline "test"

operator/pkg/controllers/etcd/statefulset.go
```diff
@@ -36,7 +36,10 @@ func (c *Controller) reconcileStatefulSet(ctx context.Context, controlPlane *v1a
 	if err != nil {
 		return fmt.Errorf("failed to patch pod spec, %w", err)
 	}
-
+	// If the custom node selector provided is "" skip adding the nodeselector to the pod
```
ellistarn, line 39: This seems like a total hack. Can we model this in another way?

ellistarn (approved):

PR #163, opened 2022-03-24
Add permission needed for tekton pipeline

felix-zhe-huang:
Issue #, if available:

Description of changes:
Currently the kitctl substrate installs Tekton without setting up the RBAC permissions needed to execute the KIT pipeline tasks.  In this change, we include new RBAC permissions and create a separate service account for the pipeline execution.
We also include new pipeline and task definitions for setting up, tearing down kit test clusters.
 
How this change was tested:
These changes were manually tested with my development environment.


substrate/pkg/controller/substrate/cluster/addons/tekton.go
```diff
@@ -43,6 +162,9 @@ func (t *Tekton) Create(ctx context.Context, substrate *v1alpha1.Substrate) (rec
 			return reconcile.Result{}, fmt.Errorf("applying tekton, %w", err)
 		}
 	}
+	if err := client.ApplyYAML(ctx, []byte(rbac)); err != nil {
```
ellistarn, line 165: Why do we need to apply this? Is the tekton release broken?

ellistarn:
Can you add a description on how this was tested and why we need this change?

substrate/pkg/controller/substrate/cluster/addons/tekton.go
felix-zhe-huang: The tekton release yaml files defines the minimum RBAC rules for it to run. However, we need a separate RBAC rules for the service account that execute the tasks in the pipeline. Those are usually defined by the Tekton user. Since Tekton now is part of KIT as add-on, it should come with the permissions to deploy control and data planes. These are the rules that it needs for the current pipeline. We will adjust it if in the future we change the control plane and data plane deployment tasks.

substrate/pkg/controller/substrate/cluster/addons/tekton.go
```diff
@@ -23,6 +23,125 @@ import (
...
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
+  name: tekton-dashboard-kit
+rules:
+- apiGroups:
+  - kit.k8s.sh
+  resources:
+  - controlplanes
+  - dataplanes
+  verbs:
```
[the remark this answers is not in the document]
felix-zhe-huang: done

substrate/pkg/controller/substrate/cluster/addons/tekton.go
```diff
@@ -23,6 +23,69 @@ import (
...
+  resources: ["signers"]
+  verbs: ["approve"]
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app.kubernetes.io/component: dashboard
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-dashboard
```
prateekgogia, line 67: For my understanding, why do we need these labels?
Why is this SA associated with dashboard?

tests/pipelines/kit/etcd-list-job-pipeline.yaml
```diff
@@ -0,0 +1,138 @@
...
+    default: "50"
+    description: Number of dummy pods
+  - name: pod-size
+    default: "5000"
+    description: Size of the pod object
+  - name: parallelism
+    default: "11"
+    description: K8S job parallelism
+  - name: etcd-list-job-image
+    default: "285627164574.dkr.ecr.us-west-2.amazonaws.com/etcd-load-tests:latest"
```
prateekgogia, line 60: I don't think we should add this particular test here as this is a very specific test cases needed by etcd team and here we want more generic tests

tests/pipelines/kit/etcd-list-job-pipeline.yaml
```diff
@@ -0,0 +1,138 @@
...
+kind: Pipeline
+metadata:
+  name: etcd-list-job-pipeline
+  namespace: tekton-pipelines
+spec:
+  params:
+  - name: name
+    description: The name of the test cluster.
+    default: "guest"
+  - name: scheduler-replicas
```
prateekgogia, line 12: Lets syncup offline about all these params, most of this should be default.

tests/tasks/generators/manual-deployment/deploy-benchmark-payload.yaml
```diff
@@ -0,0 +1,71 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: deploy-benchmark-payload
```
prateekgogia, line 5: Although we should be relying on clusterLoader2 for these tasks, but we can keep it for now and switch to clusterLoader later. 
However, we should make the task name more descriptive and clear as these tasks should clearly reflect what they are doing.

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: control-plane-setup
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/version: "0.2"
```
prateekgogia, line 8: Why is this label needed?

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
...
+  params:
+  - name: name
+    default: "guest"
+    description: Name of the guess cluster
+  - name: scheduler-replicas
+    default: "1"
+    description: Number of scheduler replicas
+  - name: controller-manager-replicas
+    default: "1"
+    description: Number of controller manager replicas
```
prateekgogia, line 27: These replicas for KCM and scheduler are not configurable as they run as daemonsets.

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
...
+  - name: apiserver-replicas
+    default: "1"
+    description: Number of APIserver replicas
+  - name: apiserver-image
+    default: "public.ecr.aws/eks-distro/kubernetes/kube-apiserver:v1.21.9-eks-1-21-11"
```
prateekgogia, line 32: Skip default value here as its already default by operator and user should provide this from pipelines based on testcase being run.

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
...
+    default: "[]"
+    description: Parameters of the apiserver container
+  - name: apiserver-instance-type
+    default: "m5.16xlarge"
+    description: Instance type for the apiserver
+  - name: etcd-replicas
+    default: "3"
+    description: Number of ETCD replicas
+  - name: etcd-image
+    default: "public.ecr.aws/eks-distro/etcd-io/etcd:v3.4.18-eks-1-21-11"
```
prateekgogia, line 44: same.

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
...
+  - name: kubernetes-version
+    default: "1.19"
+    description: Kubernetes version for the guest cluster 
+  steps:
+  - name: setup-control-plane
+    image: gcr.io/cloud-builders/kubectl@sha256:8ab94be8b2b4f3d117f02d868b39540fddd225447abf4014f7ba4765cb39f753
+    script: |
+      #!/bin/bash
+      echo "Approving KCM requests"
+      kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
```
prateekgogia, line 61: Do we need it here, we are already doing it in the end?

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,124 @@
...
+          data=$(kubectl get secret -n $(params.name) $(params.name)-kube-admin-config -ojsonpath='{.data.config}' 2>/dev/null) || true
+          if [[ ! -z "$data" ]]; then
+              echo $data | base64 -d > /tmp/kubeconfig
+              break
+          fi
+          sleep 1
+      done
+      echo "Waiting for the APIserver endpoint to be ready"
+      endpoint=$(cat /tmp/kubeconfig  | grep server | awk '{print $2}')
+      while true; do
```
prateekgogia, line 114: I feel like there should be a timeout trying this, as we don't want to hang forever.

substrate/pkg/controller/substrate/cluster/addons/tekton.go
ellistarn: These are standard kubernetes labels https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/, copied from the tekton release.

tests/pipelines/kit/etcd-list-job-pipeline.yaml
felix-zhe-huang: Agree. Please see the new version.

tests/tasks/generators/manual-deployment/deploy-benchmark-payload.yaml
felix-zhe-huang: Agree. Done.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: You are right we don't need these labels yet. Those are just to help us identify the version of the tasks. I will remove those.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: These have been removed.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: Done.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: Done.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: This will unblock Tekton dashboard from getting status update if the tekton node is dynamically provisioned by Karpenter.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: I agree. I was struggle to get a good understanding how much TTL to give it. Can we come back to this once we switch to use cluster IP instead of NLB? In that way the TTL will be more preditable.

tests/pipelines/kit/etcd-list-job-pipeline.yaml
felix-zhe-huang: agree. that's a mistake. I have removed it.

tests/pipelines/kit/upstream-load.yaml
```diff
@@ -1,120 +0,0 @@
----
```
hakuna-matatah, line 1: what is the reason for deleting this file ?

tests/tasks/setup/kit-cluster/setup-control-plane.yaml
```diff
@@ -0,0 +1,133 @@
...
+          if [[ ! -z "$data" ]]; then
+              echo $data | base64 -d > /tmp/kubeconfig
+              break
+          fi
+          sleep 1
+      done
+      echo "Waiting for the APIserver endpoint to be ready"
+      endpoint=$(cat /tmp/kubeconfig  | grep server | awk '{print $2}')
+      while true; do
+          state=$(curl -k --connect-timeou 5 -s $endpoint/healthz) 2>/dev/null || true
```
hakuna-matatah, line 124: `connect-timeout` has a typo in it.

tests/tasks/teardown/kit-cluster/teardown.yaml
```diff
@@ -0,0 +1,30 @@
+---
```
hakuna-matatah, line 1: file name stutters here. I like the teardown.yaml that i had before. Wdyt ?

tests/pipelines/kit-cluster/pipeline-template.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
hakuna-matatah, line 1: file name stutters with folder name ? Why are we doing this ?
Also, what's the point of this pipeline ? It just sets-up KIT CP, DP and tears down, am i missing something here ?

tests/tasks/setup/kit-cluster/setup-control-plane.yaml
felix-zhe-huang: Good catch! Thanks for pointing it out.

tests/pipelines/kit-cluster/run-pipeline.template.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: tekton.dev/v1beta1
```
hakuna-matatah, line 1: i don't see value in this file too given it's referring to the pipeline that i left a comment about.

tests/pipelines/kit-cluster/pipeline-template.yaml
felix-zhe-huang: This is a template and we will write a tutorial about how to extend it to run tests. I agree that the folder name stutters. I will rename it as template.yaml.

tests/pipelines/kit/upstream-load.yaml
felix-zhe-huang: So perviously we have pipeline for eksctl, this one and the new pipeline from this PR. They are separated into eksctl, kit-cluster, and kitctl folders. After some discussion we think it will be cleaning to just have one for eksctl and one for kitctl. 
I have revised it back to the three folder structure to keep all of these pipelines.

tests/tasks/teardown/kit-cluster/teardown.yaml
felix-zhe-huang: Agree. Revised accordingly.

tests/pipelines/kit-cluster/run-pipeline.template.yaml
felix-zhe-huang: This will be an example to show how Cx should setup their run.  Especially which service account to be used and what node selector to be used.

tests/pipelines/kit-cluster/pipeline-template.yaml
hakuna-matatah: can we have a more meaningful example where we do something like run a test or scale upto `n` pods when we bring up kit cluster that Cx can refer to instead of this ?

tests/pipelines/kitctl/run.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: tekton.dev/v1beta1
```
hakuna-matatah, line 1: I don't  understand your intention of having 2 different folders here like `kitctl` and `kit-cluster` ?   Those two are essentially pipelines that launch kit clusters , of-course one runs upstream test and other doesn't. I feel like template.yaml is a subset of what `upstream-load.yaml` does, right ? Having them in two different folders doesn't make sense to me. wdyt ?

tests/tasks/setup/kitctl/setup-control-plane.yaml
```diff
@@ -0,0 +1,133 @@
+---
```
hakuna-matatah, line 1: file name stutters, how about just `controlplane.yaml` or `cp.yaml` ?

tests/tasks/setup/kitctl/setup-data-plane.yaml
```diff
@@ -0,0 +1,52 @@
+---
```
hakuna-matatah, line 1: file name stutters.
How about `dp.yaml` or `dataplane.yaml `?

tests/tasks/teardown/kitctl.yaml
```diff
@@ -0,0 +1,30 @@
...
+  steps:
+  - name: teardown
+    image: bitnami/kubectl
+    script: |
+      #!/bin/bash
+      echo "Tear down guest cluster"
+      kubectl delete controlplane -n $(params.name) $(params.name)
+      echo "Delete namespace"
+      kubectl delete namespace $(params.name)
+
```
hakuna-matatah, line 29: Nit: take off these extra lines. May be your IDE settings ?

tests/tasks/generators/manual-deployment/deploy-pods-with-size.yaml
```diff
@@ -0,0 +1,69 @@
...
+  params:
+  - name: name
+    default: "guest"
+    description: Name of the guest cluster
+  - name: number-of-pods
+    default: "50"
+    description: Number of dummy pods
+  - name: pod-size
+    default: "5000"
+    description: Size of the pod object
```
hakuna-matatah, line 25: is this is MB ? or Bytes ? Helps if it's more descriptive ? wdyt ?

tests/tasks/generators/manual-deployment/deploy-pods-with-size.yaml
```diff
@@ -0,0 +1,69 @@
...
+                  - "-ec"
+                  - "echo -n __LOADSTRING__ >> /config/output.txt && while true; do sleep 100; done"
+                volumeMounts:
+                - mountPath: /config
+                  name: config
+            volumes:
+            - name: config
+              emptyDir: {}
+      EOF
+      sed -i "s/__LOADSTRING__/$(tr -dc A-Za-z0-9 < /dev/urandom | dd bs=$(params.pod-size) count=1 2>/dev/null)/g" /tmp/deploy.yaml
```
hakuna-matatah, line 65: could you add a comment here ? like what you're doing ? allocating the block size to the pod looks like ?

hakuna-matatah:
> Can you add a description on how this was tested and why we need this change?

+1

tests/pipelines/kitctl/run.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  params:
+  - name: name
+    value: "example"
+  - name: substrate-name
+    value: "YOUR_SUBSTRATE_NAME"
+  podTemplate:
+    nodeSelector:
+      kubernetes.io/arch: amd64
+  serviceAccountName: tekton-pipelines-executor
```
hakuna-matatah, line 15: i don't think this will give all you need to run this pipeline successfully. Have you tried running this template ? What was the output ?

tests/pipelines/kit-cluster/pipeline-template.yaml
felix-zhe-huang: This template is used as the minimal example to use the modules of the Tekton pipeline. We discussed about what type of getting start example we should provide. There are two items:

-  **Minimal setup with minimal parameters:** This is to demonstrate that the simplest test requires very little hand holding. 
 
-  **Modular design:** We need to demonstrate how the Tekton module works. So instead of us providing a payload like scaling upto `n` pods,  what we will do is to create a new task in the tutorial from scratch and teach them how to insert it into this pipeline template.

Therefore, we will try to keep the pipeline as bare-bone as possible

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: IIUC, the `upstream-load.yaml` is calling the `kit-cluster-create` task which has the assumption that the environment cluster is a EKS cluster created by CDK. This task will not work for an environment created by `kitctl`. 

To provide a working example for `kitctl` cluster, we need a pipeline that uses tasks that only deal with native k8s environment. 

For pipeline folders I think we can combine the `kitctl` and `kit-cluster`. But I think we should separate the tasks folders. If we want to have consistency,  then we will keep pipeline and tasks both separated in different folders.

tests/tasks/setup/kitctl/setup-control-plane.yaml
felix-zhe-huang: Agree, done.

tests/tasks/setup/kitctl/setup-data-plane.yaml
felix-zhe-huang: done.

tests/tasks/teardown/kitctl.yaml
felix-zhe-huang: done.

tests/tasks/generators/manual-deployment/deploy-pods-with-size.yaml
felix-zhe-huang: Agree. Done.

tests/tasks/generators/manual-deployment/deploy-pods-with-size.yaml
felix-zhe-huang: done.

tests/pipelines/kitctl/run.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: pipeline-template
+  namespace: tekton-pipelines
+spec:
+  params:
+  - name: name
+    value: "example"
+  - name: substrate-name
+    value: "YOUR_SUBSTRATE_NAME"
+  podTemplate:
```
ellistarn, line 12: Is this necessary?

tests/pipelines/kitctl/run.yaml
ellistarn, line 15: Is it possible to use a default?

tests/pipelines/kitctl/run.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipeline-template
+  namespace: tekton-pipelines
+spec:
+  params:
+  - name: name
+    value: "example"
+  - name: substrate-name
```
ellistarn, line 10: @prateekgogia can you help with naming

tests/pipelines/kitctl/run.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipeline-template
+  namespace: tekton-pipelines
```
ellistarn, line 5: Why are we running in this namespace? `tekton-pipelines` is the namespace where the tekton controller runs. Any reasons to not run in the default namespace?

tests/pipelines/kitctl/template.yaml
```diff
@@ -0,0 +1,36 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: pipeline-template
```
ellistarn, line 5: If we think of this in terms of a library, this name will conflict. Maybe we can come up with naming scheme like folder == pipeline name, e.g. `kitctl` here?

tests/pipelines/kitctl/template.yaml
```diff
@@ -0,0 +1,36 @@
...
+  namespace: tekton-pipelines
+spec:
+  params:
+  - name: name
```
ellistarn, line 9: Can we collapse these names?

tests/tasks/setup/kitctl/dataplane.yaml
```diff
@@ -0,0 +1,52 @@
...
+      apiVersion: kit.k8s.sh/v1alpha1
+      kind: DataPlane
+      metadata:
+        name: $(params.name)
+        namespace: $(params.name)
+      spec:
+        clusterName: $(params.name) # Associated Cluster name
+        nodeCount: $(params.node-count)
+        subnetSelector:
+          karpenter.sh/discovery: $(params.substrate-name)
```
ellistarn, line 41: It breaks encapsulation to pass the `substrate-name` concept through. Can we call this `discovery-name` or something similar? The value we pass in can be `substrate-name`.

tests/tasks/teardown/kitctl.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  description: |
+    Tear down the guest cluster.
+  params:
+  - name: name
+    default: "guest"
+    description: Name of the guess cluster
+  steps:
+  - name: teardown
+    image: bitnami/kubectl
```
ellistarn, line 22: https://hub.docker.com/r/alpine/k8s

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: This parameter is needed for the subnet selector. We can learn the subnet selector from Karpenter provisioner in the script but that requires read permission to the provisioner CRD. It will be easier to enter the name here.

Will rename this accordingly.

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: The Tekton runs use the default serviceaccount for the namespace of the run object. That account is lacking the RBAC permissions to perform tasks defined by the tasks. Rather than given strong RBAC permissions to the default account, I think will be better to create a dedicated service account with just the right amount of permission.

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: We are going to use `management cluster` instead.

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: I think we should keep all the Tekton CRs in the same namespace for consistency. What do you think?

tests/pipelines/kitctl/template.yaml
felix-zhe-huang: This is a template and they should not use `pipeline-template` as the pipeline name for the final pipeline. I think it's the customer's job to make sure that all their pipelines are uniquely named according to what they want to do. For example, if i am extending this template to test ETCD qps performance, I should name the pipeline `ETCD-QPS-TEST`.

tests/pipelines/kitctl/template.yaml
felix-zhe-huang: I have modified the data plane creation task to auto-detect the subnet and security group selector. Now we no longer need the name of the management cluster.

tests/tasks/setup/kitctl/dataplane.yaml
felix-zhe-huang: I have modified the data plane creation task to auto-detect the subnet and security group selector. Now we no longer need the name of the management cluster.

tests/tasks/teardown/kitctl.yaml
felix-zhe-huang: This bitnami image is 25% smaller than the k8s toolbox image. Shall we replace the image later if we need tools other than `kubectl`?

tests/pipelines/kitctl/run.yaml
felix-zhe-huang: This parameter is no longer needed.

tests/tasks/setup/kitctl/dataplane.yaml
```diff
@@ -0,0 +1,52 @@
...
+        nodeCount: $(params.node-count)
+        subnetSelector:
+          karpenter.sh/discovery: ${TAG}
+      EOF
+      echo "Getting kube admin config"
+      kubectl get secret -n $(params.name) $(params.name)-kube-admin-config -ojsonpath='{.data.config}' | base64 -d > /tmp/kubeconfig
+      echo "Waiting for the worker nodes to be ready"
+      while true; do
+          ready_node=$(kubectl --kubeconfig=/tmp/kubeconfig get nodes 2>/dev/null | grep -w Ready | wc -l)
+          if [[ "$ready_node" -eq $(params.node-count) ]]; then break; fi
```
hakuna-matatah, line 48: does it ever be equal even when all the nodes are ready ? I think you may get one more than the actual nodecount with `wc -l`

felix-zhe-huang: The `grep -w Ready` is filtering out the first row of the output so that only nodes that are `Ready` will be outputted.  `wc -l` will count the correct number of nodes.

prateekgogia (approved):
We can revisit some of the structuring for test definitions and namespace where all the jobs are created.

Issue #161, opened 2022-03-23
[operator] API server crashloops for version 1.21


prateekgogia:
After adding the `--insecure-port` for monitoring in 1.21 API server crashes with this error-
```
Flag --insecure-port has been deprecated, This flag has no effect now and will be removed in v1.24.
Flag --insecure-bind-address has been deprecated, This flag has no effect now and will be removed in v1.24.
Error: invalid port value 8080: only zero is allowed
```

PR #165, opened 2022-03-24
Monitor API server on secure port

prateekgogia:
Issue #, if available:
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/161

Description of changes:
- Configure pod monitors to configure prometheus scrape API server on secure port
- Upgraded kube module dependencies as importing pod monitor was causing issues.
- `client.MergeFrom` in upgraded version accepts client.Object as oppose to earlier it did `runtime.Object`. So had to implement `DeepCopy()` method.
- Create RBAC rules in guest cluster to allow `system:monitoring` group to access `/metrics` endpoint.
- Tests were failing with error `no matches for kind "PodMonitor" in version "monitoring.coreos.com/v1"`  as podmonitor CRD definition was not found in test environment cluster created. Adding the pod monitor CRDs fixes the issue, so I have included the CRD definition in environment directory until we figure a better solution for this.
 

operator/charts/kit-operator/templates/controller/rbac.yaml
```diff
@@ -88,6 +88,17 @@ rules:
   - events
   verbs:
   - create
+- apiGroups:
+  - "monitoring.coreos.com"
+  resources:
+  - podmonitors
+  verbs:
+  - get
```
ellistarn, line 96: consider collapsing these `verbs: [ ... ]`

operator/pkg/controllers/addons/rbac.go
```diff
@@ -0,0 +1,81 @@
...
+
+// RBACController configures the required RBAC permissions in the cluster for monitoring
+func RBACController(kubeClient *kubeprovider.Client) *RBAC {
+	return &RBAC{kubeClient: kubeClient}
+}
+
+type reconcileRBACResources func(context.Context) (err error)
+
+func (r *RBAC) Reconcile(ctx context.Context, _ *v1alpha1.ControlPlane) error {
+	for _, reconcile := range []reconcileRBACResources{
```
ellistarn, line 38:
```suggestion
	for _, reconcile := range []func(context.Context) (err error){
```

operator/pkg/controllers/controller.go
```diff
@@ -59,7 +58,7 @@ func (c *GenericController) Reconcile(ctx context.Context, req reconcile.Request
 		resource.GetObjectKind().SetGroupVersionKind(v1alpha1.SchemeGroupVersion.WithKind(v1alpha1.ControlPlaneKind))
 	}
 	// 2. Copy object for merge patch base
-	persisted := resource.DeepCopyObject()
+	persisted := c.DeepCopy(resource)
```
ellistarn, line 61: Why did you add this deep copy interface? It seems like you've wired it through a bunch of controllers, but are just casting. You might as well just cast here.

operator/pkg/test/environment/environment.go
```diff
@@ -55,8 +55,9 @@ func (e *Environment) Stop() error {
 	return e.Environment.Stop()
 }
 
-func crdFilePath() string {
+func crdFilePaths() []string {
 	_, file, _, _ := runtime.Caller(0)
 	p := filepath.Join(filepath.Dir(file), "..", "..", "..")
-	return filepath.Join(p, "charts/kit-operator/crds/control-plane-crd.yaml")
+	// includes CRDs for podmonitors
+	return []string{filepath.Join(p, "charts/kit-operator/crds/control-plane-crd.yaml"), filepath.Dir(file)}
```
ellistarn, line 62: It should work with just the folder, I think.

ellistarn (commented):
very cool. nice work!

operator/pkg/test/environment/environment.go
prateekgogia: Yeah the `filepath.Dir(file)` just passes the directory. In case of `charts/kit-operator/crds/control-plane-crd.yaml` we are only installing control-plane CRD and skipping the Dataplane CRD.

operator/pkg/controllers/controller.go
prateekgogia: In older versions for controller-runtime libraries `client.MergeFrom()` used to accept `runtime.Object` but in updated versions`client.MergeFrom()` accepts `client.Object`. `DeepCopyObject()` returns a `runtime.Object`.

Here we will have to cast both ControlPlane and Dataplane objects and if we at some point add more higher level objects to the controller this switch cases will grow.

ellistarn (approved):

Issue #133, opened 2022-02-23
[operator] Pending tasks for EBS support in etcd


prateekgogia:
- [ ] Update the readme to include steps to install EBS CSI driver
- [ ] PVCs are not cleaned up after deleting the control plane object
We will need to clean them in operator, until this feature is available https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/
- [ ] [Document] PVCs are not configurable for an existing cluster, user will need to delete the cluster and recreate the cluster desired PVCs
Statefulsets don't allow updating PVC spec - spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"
- [ ] Have an option in the control plane spec to let the user create etcd with host volume mounted to the pods

PR #179, opened 2022-04-08
Add controller to add ownerReference to PVC

felix-zhe-huang:
Issue #, if available:
#133 
Description of changes:
Add a controller that insert the corresponding control plane object as the owner reference of the ETCD PVCs.


operator/pkg/controllers/etcd/persistentvolumeclaim.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (c *Controller) reconcilePersistentVolumeClaims(ctx context.Context, controlPlane *v1alpha1.ControlPlane) (errs error) {
+	for i := 0; i < controlPlane.Spec.Etcd.Replicas; i++ {
+		pvc := &v1.PersistentVolumeClaim{}
+		pvcKey := types.NamespacedName{Namespace: controlPlane.Namespace, Name: fmt.Sprintf("etcd-data-%s-%d", ServiceNameFor(controlPlane.ClusterName()), i)}
+		if err := c.kubeClient.Get(ctx, pvcKey, pvc); err != nil {
+			if errors.IsNotFound(err) {
+				continue
+			}
+			errs = multierr.Append(errs, fmt.Errorf("getting pvc %s, %w", pvcKey.Name, err))
```
prateekgogia, line 37: Do we not want to `continue` after getting an err and not call Update?

operator/pkg/controllers/etcd/persistentvolumeclaim.go
```diff
@@ -0,0 +1,44 @@
+/*
```
prateekgogia, line 1: Do we want to keep PVC related stuff in one file, we already have `persistentvolumeclaimspec.go`. I would just create in one file and call `persistentvolumeclaim.go`

operator/pkg/controllers/etcd/persistentvolumeclaim.go
felix-zhe-huang: Good point. I was thinking having both error messages about failed to get PVC and failed to update.
But one error message should be good enough.

Done.

operator/pkg/controllers/etcd/persistentvolumeclaim.go
felix-zhe-huang: good point. Done.

prateekgogia (approved):
LGTM!

PR #188, opened 2022-04-18
Sync KIT flags with EKS flags

natherz97:
Issue #, if available: 98

### Description: 
This change will make the settings for apiserver, authenticator, controller-manager, scheduler, aws-encryption-provider, and etcd in KIT consistent with those used in EKS. 

Version-specific issues:
- 1.19 does not support the CSIServiceAccountToken feature gate, https://kubernetes.io/blog/2020/12/18/kubernetes-1.20-pod-impersonation-short-lived-volumes-in-csi/
- 1.22 introduces CCM in EKS. If we plan to enable CCM in KIT in K8s version X, we will need to add the leader-migration-config flag to KCM in versions X and X-1 to control the migration, https://kubernetes.io/docs/tasks/administer-cluster/controller-manager-leader-migration/

To address version-specific features in KIT in the future, we may need to build a mechanism to apply different flags per K8s version.

### Testing:
Provisioned a KIT cluster with versions 1.19, 1.20, and 1.21 using these changes.


operator/pkg/controllers/etcd/pod.go
```diff
@@ -103,6 +103,9 @@ func podSpecFor(controlPlane *v1alpha1.ControlPlane) *v1.PodSpec {
 				"--trusted-ca-file=/etc/kubernetes/pki/ca.crt",
 				"--logger=zap",
 				"--quota-backend-bytes=8589934592",
+				"--max-request-bytes=1572864",
```
hakuna-matatah, line 106: is this the max that we could today ?

operator/pkg/controllers/master/kubeapiserver.go
```diff
@@ -140,6 +140,16 @@ func apiServerPodSpecFor(controlPlane *v1alpha1.ControlPlane) v1.PodSpec {
 					"--tls-private-key-file=/etc/kubernetes/pki/apiserver/apiserver.key",
 					"--authentication-token-webhook-config-file=/var/aws-iam-authenticator/kubeconfig/kubeconfig.yaml",
 					"--encryption-provider-config=/etc/kubernetes/aws-encryption-provider/encryption-configuration.yaml",
+					"--profiling=false",
```
hakuna-matatah, line 143: As we are going to use this for testing, why turn this off ? Isn't it better for us to have it turned on for better debugging ?

operator/pkg/controllers/etcd/pod.go
natherz97: This could be increased but we're using the same as the upstream default, https://github.com/etcd-io/etcd/blob/bfb9aa42059d84ea60a4cddd75d7f3e94b27935c/server/embed/config.go#L61. It's possible to use a larger value but we'd get a warning, https://github.com/etcd-io/etcd/blob/c83b1ad9bad637fd02c6e17010755c9b35860bd7/server/etcdserver/bootstrap.go#L56

hakuna-matatah (approved):
lgtm

prateekgogia (approved):
LGTM!

prateekgogia (approved):

PR #190, opened 2022-04-18
create eks clusters and nodegroups with awscli with eks api endpoint

hakuna-matatah:
Issue #, if available:

Description of changes:
This PR creates tekton tasks to create EKS clusters and Managed nodegroups against a given eks api endpoint.


tests/tasks/setup/eks/awscli-dp.yaml
```diff
@@ -0,0 +1,87 @@
...
+  - name: validate-nodes
+    image: alpine/k8s:1.22.6
+    workingDir: $(workspaces.config.path)
+    script: |
+      mkdir -p /root/.kube/
+      cp $(workspaces.config.path)/kubeconfig /root/.kube/config
+      #kubectl commands are purely for knowing state of cluster before kicking off the test.
+      kubectl version
+      kubectl config current-context
+      kubectl describe clusterrole eks:node-manager
```
prateekgogia, line 80: Isn't this static, why do we need this?

tests/tasks/setup/eks/awscli-dp.yaml
```diff
@@ -0,0 +1,87 @@
...
+      kubectl get nodes -o wide
+      kubectl get ns
+      while true; do
+          ready_node=$(kubectl get nodes 2>/dev/null | grep -w Ready | wc -l )
+          if [[ "$ready_node" -eq $(params.desired-nodes) ]]; then break; fi
+          sleep 10
```
prateekgogia, line 86: In the worst case we would wait for 10 seconds after all nodes are ready. Can we reduce it 2 seconds? I guess its alright to check once every couple of seconds

prateekgogia (commented):
Can we add pipeline and pipelinerun spec for these tasks?
Are you also planning to push the change for IAM permissions as part of this PR?

tests/tasks/setup/eks/awscli-cp.yaml
```diff
@@ -0,0 +1,39 @@
...
+    description: servicerole arn to be used for eksc cluster
+  workspaces:
+  - name: config
+    description: |
+      A workspace into which a kubeconfig file called `kubeconfig` will be written that will contain the information required to access the cluster. The `kubeconfig` will expect to use [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator/) to authenticate, so in order for it to be used it must be run in a container which contains both `kubectl` and `aws-iam-authenticator`.
+  steps:
+  - name: write-kubeconfig
+    image: amazon/aws-cli
+    script: |
+      aws eks create-cluster --name $(params.cluster-name) --kubernetes-version $(params.kubernetes-version) --role-arn $(params.servicerole) --resources-vpc-config subnetIds=$(params.subnets),securityGroupIds=$(params.securitygroups) --endpoint $(params.endpoint)
```
felix-zhe-huang, line 36: Can we default the endpoint to empty and only add `--endpoint $(params.endpoint)` when a non-empty endpoint URL is given?

tests/tasks/setup/eks/awscli-cp.yaml
```diff
@@ -0,0 +1,39 @@
...
+      aws eks --endpoint $(params.endpoint) wait cluster-active --name $(params.cluster-name)
+      aws eks --endpoint $(params.endpoint) update-kubeconfig --name $(params.cluster-name) --region $(params.region)
```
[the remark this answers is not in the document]
felix-zhe-huang: Can we also either print out the kubeconfig or give an example how to set up the pipeline runs that allow us to extract the kubeconfig to our laptop's filesystem? Will be nice to have access directly from laptop as well.

tests/tasks/setup/eks/awscli-dp.yaml
```diff
@@ -0,0 +1,87 @@
...
+    description: The desired number of nodes in the cluster.
+  - name: min-nodes
+    default: "1"
+    description: The minimum number of nodes in the cluster.
+  - name: max-nodes
+    default: "10"
+    description: The maximum number of nodes in the cluster.
+  - name: endpoint
+    default: "https://api.wesley.amazonaws.com"
+  - name: host-cluster-node-role-arn
```
felix-zhe-huang, line 28: Do we need to also provide a default value `tenant-controlplane-node-role` here?

tests/tasks/setup/eks/awscli-dp.yaml
```diff
@@ -0,0 +1,87 @@
...
+  workspaces:
+  - name: config
+    description: |
+      A workspace into which a kubeconfig file called `kubeconfig` will be written that will contain the information required to access the cluster. The `kubeconfig` will expect to use [aws-iam-authenticator](https://github.com/kubernetes-sigs/aws-iam-authenticator/) to authenticate, so in order for it to be used it must be run in a container which contains both `kubectl` and `aws-iam-authenticator`.
+  steps:
+  - name: create-nodegroup
+    image: amazon/aws-cli
+    workingDir: $(workspaces.config.path)
+    script: |
+      ssh_key_name=$(params.region)-nodegroups-ssh-key
```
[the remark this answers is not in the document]
felix-zhe-huang: I think this is the EC2 key pair for creating EKS cluster at line 60.

tests/tasks/setup/eks/awscli-dp.yaml
```diff
@@ -0,0 +1,87 @@
...
+      --subnets $NG_SUBNETS
+
+      while [[ "$(aws eks --endpoint $(params.endpoint) --region $(params.region) describe-nodegroup --cluster-name $(params.cluster-name) --nodegroup-name $NG_NAME --query nodegroup.status --output text)" == "CREATING" ]]
+      do
+      echo "$NG_NAME is "CREATING" at $(date)"
+      sleep 10
+      done
+      aws eks --endpoint $(params.endpoint) update-kubeconfig --name $(params.cluster-name) --region $(params.region)
+      cp  /root/.kube/config $(workspaces.config.path)/kubeconfig
+  - name: validate-nodes
```
felix-zhe-huang, line 71: Do we need to separate this into two steps? It will introduce extra overhead for spinning up new container.

tests/tasks/setup/eks/awscli-cp.yaml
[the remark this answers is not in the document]
prateekgogia: I think we will need to decide what happens when a test fails? 
Do we keep the cluster around for debugging or delete the cluster after every test run irrespective of the test result?

tests/tasks/setup/eks/awscli-dp.yaml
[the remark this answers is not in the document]
prateekgogia: Yes but we don't need SSH configured on these nodes. We can use SSM.

tests/tasks/setup/eks/awscli-dp.yaml
prateekgogia: The overhead should be very low as it's another pod, unless I am missing something?
This will give us an individual task to validate nodes joining and later we can enhance to check time taken for nodes to join or some additional checks that can pass/fail the test.

hakuna-matatah:
closed in favor of a bigger scope and different goal for eks specific tests

Issue #218, opened 2022-05-19
[kitcli] kitctl doesn't handle s3 bucket creation in us-east-1


liko9:
When using kitctl in us-east-1, I received the following error message:
`reconciling cluster.Config, err ensuring S3 bucket, creating S3 bucket, InvalidLocationConstraint: The specified location-constraint is not valid`
I've found some issues logged against the javascript AWS SDK regarding this, but nothing for the go SDK.  It seems that if the region is us-east-1 and that is default for the SDK, it will give you this error:  https://github.com/aws/aws-sdk-js/issues/3647

PR #219, opened 2022-05-19
Fix S3 creation in us-east-1

prateekgogia:
Issue #, if available:
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/218

Description of changes:
- Some regions like `us-east-1` limit default CIDR count to 5, so we need to keep the defaults
- S3 bucket creation fix in `us-east-1`


hakuna-matatah (approved):
lgtm

PR #229, opened 2022-06-10
delete nodegroups as part of cluster deletion; abort cluster creation if cluster already exist…

hakuna-matatah:
This PR adds the ability to:
-  delete all nodegroups at the time of deletion of cluster tekton task. 
- Adds `delete cluster` as a tekton `finally` step to pipeline which will help `tear` down cluster irrespective of the success/failure of previous steps in pipeline

Issue #, if available:

Description of changes:

### DIFF
--- a/tests/tasks/teardown/awscli-eks.yaml
+++ b/tests/tasks/teardown/awscli-eks.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   description: |
     Teardown an EKS cluster.
-    This Task can be used to teardown an EKS cluster in an AWS account.
+    This Task can be used to teardown an EKS cluster with mng in an AWS account.
   params:
   - name: cluster-name
     description: The name of the EKS cluster which will be teared down.
@@ -26,4 +26,10 @@ spec:
       if [ -n "$(params.endpoint)" ]; then
         ENDPOINT_FLAG="--endpoint $(params.endpoint)"
       fi
+
+      for i in `aws eks list-nodegroups --cluster-name $(params.cluster-name) $ENDPOINT_FLAG | jq -r '.nodegroups[]'`;
+      do
+          aws eks delete-nodegroup --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
+          aws eks wait nodegroup-deleted --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
+      done;
       aws eks delete-cluster --name $(params.cluster-name) --region $(params.region) $ENDPOINT_FLAG
--- a/tests/pipelines/eks/awscli-cl2-load.yaml
+++ b/tests/pipelines/eks/awscli-cl2-load.yaml
@@ -68,6 +68,16 @@ spec:
       workspace: config
     - name: results
       workspace: results
+  finally:    
+  - name: teardown
+    params:   
+    - name: cluster-name
+      value: $(params.cluster-name)
+    - name: endpoint
+      value: $(params.endpoint)
+    taskRef:
+      kind: Task
+      name:  awscli-eks-cluster-teardown
   workspaces:
   - name: config
   - name: source


tests/tasks/teardown/awscli-eks.yaml
```diff
@@ -26,4 +26,11 @@ spec:
...
-      aws eks delete-cluster --name $(params.cluster-name) --region $(params.region) $ENDPOINT_FLAG
\ No newline at end of file
+
+      for i in `aws eks list-nodegroups --cluster-name $(params.cluster-name) $ENDPOINT_FLAG | jq -r '.nodegroups[]'`;
+      do
+          aws eks delete-nodegroup --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
+          aws eks wait nodegroup-deleted --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
+      done;
+      aws eks delete-cluster --name $(params.cluster-name) --region $(params.region) $ENDPOINT_FLAG
+      done;
```
prateekgogia, line 36: I think this `done` is extra here

tests/tasks/setup/eks/awscli-cp.yaml
```diff
@@ -47,6 +47,9 @@ spec:
       
       if [ "$CREATED_CLUSTER" == "" ]; then
         aws eks create-cluster --name $(params.cluster-name) --region $(params.region) --kubernetes-version $(params.kubernetes-version) --role-arn $(params.servicerole) --resources-vpc-config subnetIds=$subnets,securityGroupIds=$sg $ENDPOINT_FLAG
+      else 
+        echo "cluster already exists; Aborting as cluster was not cleaned up already"
+        exit 1
```
prateekgogia, line 52: I don't think this is a good approach, we have a task to create a cluster with a name and if a cluster already exists it shouldn't fail just exit gracefully. 
This helps with testing and reiterating stuff as the cluster already exists and we can run tests again on this cluster.

prateekgogia (changes requested):

prateekgogia (approved):

tests/tasks/teardown/awscli-eks.yaml
```diff
@@ -26,4 +26,10 @@ spec:
       if [ -n "$(params.endpoint)" ]; then
         ENDPOINT_FLAG="--endpoint $(params.endpoint)"
       fi
+
+      for i in `aws eks list-nodegroups --cluster-name $(params.cluster-name) $ENDPOINT_FLAG | jq -r '.nodegroups[]'`;
+      do
+          aws eks delete-nodegroup --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
+          aws eks wait nodegroup-deleted --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG;
```
shyamjvs, line 33: This may be good enough for now. But sometimes these calls may fail (for e.g timeout while waiting for nodegroup to go down). What's the behavior in that case? Will we drop the `finally` block after single attempt and go ahead with the next run anyway?

If there's an "initialization" or "pre-flight check" option, might be worth adding that too as a blocker before actually starting the test logic.

hakuna-matatah: Yes, there is a possibility where Finally step can fail during server side (eks api) failures. Probability of such failures are less, and I have thought about that corner case, our defense to that failures is we are always creating clusters with same `name`, so we can only have one cluster per kit environment at any given point in time unlike before because, EKS doesn't allow user to create cluster with same `name` twice in same region. However we may want to be notified when we see server side failures during  such runs, which is why I have added this backlog issue for us  - https://github.com/awslabs/kubernetes-iteration-toolkit/issues/230

prateekgogia: Also, if we fail in the `finally` call and leak resources, in the next iteration we will reconcile on an existing cluster with the cluster creation task and `finally` will try to clean again. As Harish mentioned we will need to be notified when `finally` fails, so we need an alarm.

PR #231, opened 2022-06-13
Increase sleep time to finish VAS scale up fully

hakuna-matatah:
Issue #, if available:

Description of changes:
Increase sleep time further to finish VAS scale up fully before user tests kicks in. 

### DIFF
--- a/tests/tasks/setup/eks/awscli-mng.yaml
+++ b/tests/tasks/setup/eks/awscli-mng.yaml
@@ -38,7 +38,7 @@ spec:
       if [ $(params.desired-nodes) -lt 5001  ] && [ $(params.desired-nodes) -gt 500 ]
       then
         echo "sleeping for 3hrs to workaround VAS cool off time"
-        sleep 10800
+        sleep 12000
       fi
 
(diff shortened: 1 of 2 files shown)


tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       if [ $(params.desired-nodes) -lt 5001  ] && [ $(params.desired-nodes) -gt 500 ]
       then
         echo "sleeping for 3hrs to workaround VAS cool off time"
-        sleep 10800
+        sleep 12000
```
prateekgogia, line 41: Why do we need extra cool down period, I thought it was 3 hours?

hakuna-matatah: 3 hrs is for VAS cool off time. Nodegroups are created right after the cooloff and tests are being kicked off at that point. It is racing with VAS Scaling updates at that time, which we don't want. So, adding some extra buffer time by then VAS Scale up will be actually finished before we start doing any dataplane stuff.

prateekgogia (approved):

PR #233, opened 2022-06-17
[operator] Allow Dataplane to use Public Subnets

rschalo:
### Description of changes:
This PR adds the ability for dataplane nodes to join a guest cluster using public subnets if there are not private subnets available. This change also prevents nodes from attempting to join the guest cluster with a subnet that does not have any available IPs. This was tested with a subnet that has a 10.0.0.0/28 IPv4 CIDR.

#### Logs from `kit-controller`when no IPs available
```
2022-06-17T20:35:16.548Z	ERROR	controller.dataplane	Reconciler error	{"reconciler group": "kit.k8s.sh", "reconciler kind": "DataPlane", "name": "ip-testing-nodes", "namespace": "default", "error": "reconciling resource, creating auto scaling group for ip-testing, failed to find subnets for dataplane ip-testing-nodes"}
```
#### Logs from `kit-controller` when IPs available
```
2022-06-17T21:07:20.359Z	INFO	[regression-testing] control plane reconciled
2022-06-17T21:07:20.508Z	INFO	[regression-testing] Addons reconciled
2022-06-17T21:07:39.063Z	INFO	[regression-testing] data plane reconciled
```

### DIFF
--- a/operator/pkg/awsprovider/instances/reconciler.go
+++ b/operator/pkg/awsprovider/instances/reconciler.go
@@ -123,10 +123,10 @@ func (c *Controller) updateAutoScalingGroup(ctx context.Context, dataplane *v1al
 func (c *Controller) createAutoScalingGroup(ctx context.Context, dataplane *v1alpha1.DataPlane) error {
 	subnets, err := c.subnetsFor(ctx, dataplane)
 	if err != nil {
-		return fmt.Errorf("getting private subnet for %s, %w", dataplane.Spec.ClusterName, err)
+		return fmt.Errorf("getting subnet for %s, %w", dataplane.Spec.ClusterName, err)
 	}
 	if len(subnets) == 0 {
-		return fmt.Errorf("failed to find private subnets for dataplane")
+		return fmt.Errorf("failed to find subnets for dataplane %s", dataplane.Name)
 	}
 	_, err = c.autoscaling.CreateAutoScalingGroupWithContext(ctx, &autoscaling.CreateAutoScalingGroupInput{
 		AutoScalingGroupName: ptr.String(AutoScalingGroupNameFor(dataplane)),
@@ -180,24 +180,36 @@ func (c *Controller) subnetsFor(ctx context.Context, dataplane *v1alpha1.DataPla
 	if err != nil {
 		return nil, fmt.Errorf("getting subnet for %s, %w", dataplane.Spec.ClusterName, err)
 	}
-	return c.filterPrivateSubnets(ctx, subnetIDs)
+	// Returns public subnets if no private subnets are found
+	return c.filterSubnets(ctx, subnetIDs)
 }
 
-func (c *Controller) filterPrivateSubnets(ctx context.Context, ids []*string) ([]string, error) {
+func (c *Controller) filterSubnets(ctx context.Context, ids []*string) ([]string, error) {
+	if len(ids) == 0 {
+		return []string{}, fmt.Errorf("found zero subnets while filtering")
+	}
 	output, err := c.ec2api.DescribeSubnetsWithContext(ctx, &ec2.DescribeSubnetsInput{
 		SubnetIds: ids,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describing subnet, %w", err)
 	}
-	result := []string{}
+	public := []string{}
+	private := []string{}
 	for _, subnet := range output.Subnets {
-		if ptr.BoolValue(subnet.MapPublicIpOnLaunch) {
+		if aws.Int64Value(subnet.AvailableIpAddressCount) == 0 {
 			continue
 		}
-		result = append(result, *subnet.SubnetId)
+		if ptr.BoolValue(subnet.MapPublicIpOnLaunch) {
+			public = append(public, *subnet.SubnetId)
+		} else {
+			private = append(private, *subnet.SubnetId)
+		}
 	}
-	return result, nil
+	if len(private) == 0 {
+		return public, nil
+	}
+	return private, nil
 }
 
 func (c *Controller) subnetsForSelector(ctx context.Context, selector map[string]string) ([]string, error) {
--- a/docs/how-to-use-kit.md
+++ b/docs/how-to-use-kit.md
@@ -67,6 +67,7 @@ Add tasks and pipelines from the KIT repo, (we have a [task](https://github.com/
 How to collect metrics
 
 * Access the Grafana dashboard on [http://localhost:8080](http://localhost:8080/), make sure you have run port-forward command mentioned above.
+* After reaching the login screen at `http://localhost:8080/login`, log in with username: `admin`, password: `prom-operator`.
 * All the core Kubernetes components like API Server, KCM, scheduler, etcd, authenticator are being scraped and their metrics are being plotted in their own Grafana dashboards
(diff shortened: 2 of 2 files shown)


operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -180,24 +180,36 @@ func (c *Controller) subnetsFor(ctx context.Context, dataplane *v1alpha1.DataPla
...
 	}
-	return c.filterPrivateSubnets(ctx, subnetIDs)
+	// Returns public subnets if no private subnets are found
+	return c.filterSubnets(ctx, subnetIDs)
 }
 
-func (c *Controller) filterPrivateSubnets(ctx context.Context, ids []*string) ([]string, error) {
+func (c *Controller) filterSubnets(ctx context.Context, ids []*string) ([]string, error) {
+	if len(ids) == 0 {
+		return []string{}, fmt.Errorf("found zero subnets while filtering")
```
prateekgogia, line 189: nit: 
```suggestion
		return nil, fmt.Errorf("found zero subnets while filtering")
```

prateekgogia (dismissed):
LGTM, nice work!

operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -123,10 +123,10 @@ func (c *Controller) updateAutoScalingGroup(ctx context.Context, dataplane *v1al
 func (c *Controller) createAutoScalingGroup(ctx context.Context, dataplane *v1alpha1.DataPlane) error {
 	subnets, err := c.subnetsFor(ctx, dataplane)
 	if err != nil {
-		return fmt.Errorf("getting private subnet for %s, %w", dataplane.Spec.ClusterName, err)
+		return fmt.Errorf("getting subnet for %s, %w", dataplane.Spec.ClusterName, err)
 	}
 	if len(subnets) == 0 {
-		return fmt.Errorf("failed to find private subnets for dataplane")
+		return fmt.Errorf("failed to find subnets for dataplane %s", dataplane.Name)
```
ellistarn, line 129: It would be great if the logger included context such that you didn't need to annotate `dataplane %s", dataplane.Name)`. For example, if at the top level, your logger did something like .With("dataplane", dataplane.Name"), this would magically wire up for all log lines. We do this broadly across karpenter. 

https://github.com/aws/karpenter/blob/main/pkg/controllers/provisioning/provisioner.go#L135

operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -180,24 +180,36 @@ func (c *Controller) subnetsFor(ctx context.Context, dataplane *v1alpha1.DataPla
...
 		return nil, fmt.Errorf("getting subnet for %s, %w", dataplane.Spec.ClusterName, err)
 	}
-	return c.filterPrivateSubnets(ctx, subnetIDs)
+	// Returns public subnets if no private subnets are found
+	return c.filterSubnets(ctx, subnetIDs)
 }
 
-func (c *Controller) filterPrivateSubnets(ctx context.Context, ids []*string) ([]string, error) {
+func (c *Controller) filterSubnets(ctx context.Context, ids []*string) ([]string, error) {
+	if len(ids) == 0 {
```
ellistarn, line 188: It looks like theres some redundancy with the emptiness checking logic. 

Previously this filter function would simply return [] if no subnets were found, rather than erroring. [Callers](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/233/files#diff-295f7552e000471288f85f68537af958d7074254753d9a3935a85bde73db8ac9R124) were responsible for behaving correctly if there weren't any subnets. 

It's important to differentiate between transient network errors (i.e. retriable) and permanent logical errors (e.g. no subnets configured). If you comingle these concepts, you risk your code failing rarely for unknown reasons.

operator/pkg/awsprovider/instances/reconciler.go
prateekgogia: yeah, I like the idea. 
This might be a bigger change, not blocking this PR for it.

prateekgogia (approved):

operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -180,24 +180,36 @@ func (c *Controller) subnetsFor(ctx context.Context, dataplane *v1alpha1.DataPla
...
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describing subnet, %w", err)
 	}
-	result := []string{}
+	public := []string{}
+	private := []string{}
 	for _, subnet := range output.Subnets {
-		if ptr.BoolValue(subnet.MapPublicIpOnLaunch) {
+		if aws.Int64Value(subnet.AvailableIpAddressCount) == 0 {
```
ellistarn, line 200: Both aws/aws-sdk-go and knative/pkg have libraries for inverting and reverting pointers like this. It looks like you're swapping from relying on knative pkg to the aws sdk. But then on line 203, you're using knative/pkg again.

operator/pkg/awsprovider/instances/reconciler.go
```diff
@@ -180,24 +180,36 @@ func (c *Controller) subnetsFor(ctx context.Context, dataplane *v1alpha1.DataPla
...
 	output, err := c.ec2api.DescribeSubnetsWithContext(ctx, &ec2.DescribeSubnetsInput{
 		SubnetIds: ids,
 	})
 	if err != nil {
 		return nil, fmt.Errorf("describing subnet, %w", err)
 	}
-	result := []string{}
+	public := []string{}
+	private := []string{}
 	for _, subnet := range output.Subnets {
```
ellistarn, line 199: Check out https://github.com/samber/lo. 

```
public := lo.Filter(subnets, func(subnet *ec2.Subnet, _ int) { return ptr.BoolValue(subnet.MapPublicIpOnLaunch) })
private := lo.Reject(subnets, func(subnet *ec2.Subnet, _ int) { return ptr.BoolValue(subnet.MapPublicIpOnLaunch) })
if len(private) > 0 {
  return private
}
return public
```

ellistarn: Is it ever the case that we might want to mix and match public and private subnets?

operator/pkg/awsprovider/instances/reconciler.go
rschalo: Choosing to go with knative/pkg

operator/pkg/awsprovider/instances/reconciler.go
rschalo: If private subnets are available they are preferred.

rschalo: @prateekgogia do we want to take the dependency on `samber/lo`?

operator/pkg/awsprovider/instances/reconciler.go
rschalo: Changed to the empty slice return per our conversation on Friday.

operator/pkg/awsprovider/instances/reconciler.go
prateekgogia: Usually, I take a dependency on other modules if we have logic thats being used at more than 1 place / can be error prone. 
In this particular case, I think its straightforward and easy to understand what you are trying to do. IMO, we can keep at it as-is.

prateekgogia (approved):

PR #260, opened 2022-08-02
add prometheus crds and fix kit permission

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Add prometheus podmonitor CRD 
 - Fix security group permission for KIT
 - Add test namespace 

### DIFF
--- a/infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
+++ b/infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
@@ -1,0 +1,37 @@
+apiVersion: karpenter.k8s.aws/v1alpha1
+kind: AWSNodeTemplate
+metadata: 
+  name: kit
+spec:
+  amiFamily: AL2
+  subnetSelector:
+    kit.sh/stack: KITInfrastructure
+    aws-cdk:subnet-type: Private
+  securityGroupSelector:
+    kit.sh/stack: KITInfrastructure
+---
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata: 
+  name: kit
+spec: 
+  requirements: 
+  - key: karpenter.sh/capacity-type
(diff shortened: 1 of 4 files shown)


infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
```diff
@@ -0,0 +1,37 @@
...
+  subnetSelector:
+    kit.sh/stack: KITInfrastructure
+    aws-cdk:subnet-type: Private
+  securityGroupSelector:
+    kit.sh/stack: KITInfrastructure
+---
+apiVersion: karpenter.sh/v1alpha5
+kind: Provisioner
+metadata: 
+  name: kit
```
prateekgogia, line 16: Why do we need another provisioner? Can't we extend the existing [default provisioner](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml)?

Having a KIT provisioner, does it mean we will need to update the pods for guest clusters which rely today on default provisioner to provision nodes?

prateekgogia: Had an offline discussion, we don't need to update guest cluster pods to provision nodes using KIT provisioner with the latest Karpenter version.

prateekgogia (approved):
LGTM!
Next task will be to upgrade karpenter installed as part of kitctl. Also use taints and tolerations for KIT pods to only be scheduled on dedicated KIT nodes.

infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
hakuna-matatah: Then what is the purpose of this provisioner being here ?

PR #263, opened 2022-08-11
Addition of shortNames to CRDs

rschalo:
Description of changes:

Generally update CRDs and add shortNames to controlplane and dataplane.


operator/charts/kit-operator/crds/control-plane-crd.yaml
```diff
@@ -12,6 +12,8 @@ spec:
     listKind: ControlPlaneList
```
prateekgogia, line 12: These files are auto generated by `hack/codegen.sh` we shouldn't be making changes directly to these files. 

We should add the `kubebuilder` [tags for ControlPlane](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/operator/pkg/apis/controlplane/v1alpha1/controlplane.go#L24) and DataPlane object definitions for shortname and run `make codegen` that will update these files.

prateekgogia (changes requested):

prateekgogia (approved):
LGTM!

PR #271, opened 2022-08-17
Fix flags for KCM in 1.22 version

prateekgogia:
Issue #, if available:

Description of changes:

Fixes couple of broken flags in KCM for 1.22 version for guest clusters.

### DIFF
--- a/operator/pkg/controllers/master/kubecontrollermanager.go
+++ b/operator/pkg/controllers/master/kubecontrollermanager.go
@@ -17,6 +17,7 @@ package master
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/awslabs/kubernetes-iteration-toolkit/operator/pkg/apis/controlplane/v1alpha1"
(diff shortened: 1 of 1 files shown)


hakuna-matatah:
Haha, these functions will be more and more as we support more and more  older versions of K8s as new versions get released upstream.  Unfortunately we are stuck with this currently but I encourage us to move in the direction of cutting release branches like upstream and not support these kind of functions in Head, if possible.

hakuna-matatah (approved):

prateekgogia:
> Haha, these functions will be more and more as we support older versions of K8s as new versions get released upstream. Unfortunately we are stuck with this currently but I encourage us to move in the direction of cutting release branches like upstream and not support these kind of functions in Head, if possible.

How will that look when we want to create different Kubernetes versions? As these will be different binaries for operator based on Kubernetes versions.

hakuna-matatah:
> > Haha, these functions will be more and more as we support older versions of K8s as new versions get released upstream. Unfortunately we are stuck with this currently but I encourage us to move in the direction of cutting release branches like upstream and not support these kind of functions in Head, if possible.
> 
> How will that look when we want to create different Kubernetes versions? As these will be different binaries for operator based on Kubernetes versions.

Well I was saying same binary for upstream supported versions of k8s and not having to maintain same binary for non-supported upstream versions of K8s, so that if someone wants to test something on older versions they can surely do so by building/downloading that release binary and do the testing.

PR #275, opened 2022-08-18
selectively disable addons

hakuna-matatah:
Issue #, if available:
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/253

Description of changes:

Selectively disables addons from Infrastructure code at the time of deployment:

### Testing

-  cdk deploy KITInfrastructure --no-rollback -c TestNamespace="tekton-pipelines" -c TestServiceAccount="tekton-pipelines-executor" -c AWSEBSCSIDriverAddon="" -c KarpenterAddon="" -c KITAddon=""

^^^ Empty values means disable addons - typescripty thing for String to Boolean.


prateekgogia:
Can we update the Readme under infrastructure as well?

infrastructure/lib/kit-infrastructure.ts
```diff
@@ -15,6 +15,9 @@ export class KITInfrastructure extends Stack {
     const repoUrl = this.getContextOrDefault('FluxRepoURL', "https://github.com/awslabs/kubernetes-iteration-toolkit")
     const repoBranch = this.getContextOrDefault('FluxRepoBranch', 'main')
     const repoPath = this.getContextOrDefault('FluxRepoPath', './infrastructure/k8s-config/clusters/kit-infrastructure')
+    const installEBSCSIDriverAddon = Boolean(this.getContextOrDefault("AWSEBSCSIDriverAddon", "true"))
```
prateekgogia, line 18: If you change this method to be `getContextOrDefault(key: string, def: string | boolean | null)` it should return the context passed and have an if condition like this - 
`if (installKitAddon == true)`

prateekgogia (changes requested):

bwagner5:
I'm curious why we'd ever want to disable those add-ons?  This setup should ideally be a very opinionated testing environment. Is there a use-case for testing without some add-ons installed?

hakuna-matatah:
> I'm curious why we'd ever want to disable those add-ons? This setup should ideally be a very opinionated testing environment. Is there a use-case for testing without some add-ons installed?

Yes, we have such use-case in Scalability where we only spin-up  EKS specific clusters through Tekton and not having to do anything with Karpenter, KIT on management cluster/KIT Infra.
These additional addons are not used at this point, in future if we choose to test KIT Guest clusters on the same infrastructure then yes, we can turn these selectively like  installing KIT addon at that point.

infrastructure/lib/kit-infrastructure.ts
hakuna-matatah: >If you change this method to be getContextOrDefault(key: string, def: string | boolean | null) it should return the context passed

It already returns the context passed with the change i made, no need to change ^^^ that function :D. Unless you are implying something here ?

>if (installKitAddon == true)

- hmmm ^^^  setting this value makes the value comparison to be explicit instead of implicit.

- IOW,  this change will ensure  any value that is set through context is taken as `false` unless it is set to `true` `explicitly`(opposite of default typescript behavior for boolean/string conversion). Instead of just empty value is taken as `false` and any other value is taken as `true`( which is default typescript behavior).

prateekgogia (approved):
LGTM!

Issue #252, opened 2022-08-02
Test the current state of KIT Management cluster with EKS


hakuna-matatah:
Test the current state of KIT Management cluster with EKS

Issue #279, opened 2022-08-22
set up CW logging for KITInfra


hakuna-matatah:
addon - to set up CW logging for KITInfra for all pods on infra cluster.

PR #280, opened 2022-08-22
Scalability infra and tests fixes

hakuna-matatah:
Issue #, if available:
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/252
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/253


infrastructure/lib/addons/aws-lbc.ts
```diff
@@ -44,6 +44,9 @@ export class AWSLoadBalancerController extends Construct {
             'eks.amazonaws.com/role-arn': sa.role.roleArn
           }
         },
+        clusterSecretsPermissions: {
+          allowAllSecrets: true
```
felix-zhe-huang, line 48: Can we configuring namespaced role/rolebinding instead of granting access to all secrets?

infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml
```diff
@@ -30,4 +30,4 @@ spec:
       memory: 1000Gi
   providerRef: 
       name: default
-  ttlSecondsAfterEmpty: 300
+  ttlSecondsAfterEmpty: 864000
```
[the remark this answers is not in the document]
hakuna-matatah: Currently we need the logs/information from stdout until we have CW agent addon is set up.

infrastructure/lib/addons/aws-lbc.ts
hakuna-matatah: These are being installed using helm charts from this repository directly - https://github.com/aws/eks-charts/tree/master/stable/aws-load-balancer-controller.
So, it already installs role/role-binding from these charts programatically without access to secrets though starting from 2.4.2 version.
Also these secrets can be in any namespace today as per Infra setup.

infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml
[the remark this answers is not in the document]
hakuna-matatah: 10 days is good enough, beyond that there isn't much use for this node to stick around.
here is the backlog ticket for that - https://github.com/awslabs/kubernetes-iteration-toolkit/issues/279

infrastructure/lib/addons/aws-lbc.ts
felix-zhe-huang: I was reading that helm chart and it is recommended to limit the scope by adding extra role/rolebinding. 
```
# clusterSecretsPermissions lets you configure RBAC permissions for secret resources
# Access to secrets resource is required only if you use the OIDC feature, and instead of
# enabling access to all secrets, we recommend configuring namespaced role/rolebinding.
# This option is for backwards compatibility only, and will potentially be deprecated in future.
clusterSecretsPermissions:
  # allowAllSecrets allows the controller to access all secrets in the cluster.
  # This is to get backwards compatible behavior, but *NOT* recommended for security reasons
  allowAllSecrets: false
```

hakuna-matatah: Yeah but in our case the scope of creation of ingress resources can be in any namespace. So, limiting it to one namespace would not work for us. Also, today the values.yaml doesn't provide us the way to do it for a single namespace. It has to be done outside of helm installation even if we choose to do it.

felix-zhe-huang (approved):
LGTM

PR #283, opened 2022-08-30
fix kit infras timeouts and block device mappings

bwagner5:
Issue #, if available:
N/A

Description of changes:
 - Fixes KIT infra timeouts (which was accidentally reverted in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/280/files) 
 - Add block device mappings so that disks don't fill up from the default 20Gi

### DIFF
--- a/infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml
+++ b/infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml
@@ -9,6 +9,11 @@ spec:
     aws-cdk:subnet-type: Private
   securityGroupSelector:
     kit.sh/stack: KITInfrastructure
+  blockDeviceMappings:
+  - deviceName: /dev/xvda
+    ebs:
+      volumeSize: 100Gi
+      volumeType: gp3
 ---
 apiVersion: karpenter.sh/v1alpha5
 kind: Provisioner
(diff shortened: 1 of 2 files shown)


infrastructure/k8s-config/clusters/kit-infrastructure/provisioner.yaml
```diff
@@ -9,6 +9,11 @@ spec:
     aws-cdk:subnet-type: Private
   securityGroupSelector:
     kit.sh/stack: KITInfrastructure
+  blockDeviceMappings:
+  - deviceName: /dev/xvda
+    ebs:
+      volumeSize: 100Gi
```
hakuna-matatah, line 15: do we really need this much memory ? Thinking in terms of Frugality principle for customers using this Infra.

What's causing disks to fill up ? If it's logs, we should probably have the CW logging enabled so we rotate the logs on the instances so disks won't fill up.

bwagner5: 100Gi is still pretty small. This is a mixture of container images that are pulled and logs for tests. We'll definitely rotate logs soon so that we don't breach this limit, but I think 100Gi is still pretty frugal but much more reasonable for the type of work the nodes are facilitating.

hakuna-matatah: ok, we can go ahead with this change for now but we should be addressing the root of the problem before we increase it further in future.
 My concern was that 100Gi won't be enough if root of the problem is not addressed. If it's log rotation, that should be addressed is what I was trying to convey. I had a backlog ticket cut for this in the past - https://github.com/awslabs/kubernetes-iteration-toolkit/issues/279 which should help flushing logs from nodes to CW.

hakuna-matatah:
LGTM!

hakuna-matatah (approved):

Issue #287, opened 2022-09-06
Make cl2 tekton task tie to perf-tests/clusterloader2 repo release branches 


hakuna-matatah:
- Today we built cl2 image for our tests but the image is not tied to any release-branch of cl2 repo - for example for k8s 1.23 release branch is -  https://github.com/kubernetes/perf-tests/tree/release-1.23 

- This causes a problem when trying to use the cl2 test image in tekton task with latest changes/config files of cl2 repo. IOW cl2 test image that was built and used in tests may not  be compatible with Head of this repo  - https://github.com/kubernetes/perf-tests and could result in unforeseen test failures 

- To avoid such issues ^^^ - idea is to tie the cl2 image used in tests with release branches of perf-tests/cl2 repo to avoid failures that could be caused due to backward incompatible changes to cl2 repo.

PR #288, opened 2022-09-06
Make cl2 tekton task tie to perf-tests/clusterloader2 repo release branches

hakuna-matatah:
Issue #, if available:
https://github.com/awslabs/kubernetes-iteration-toolkit/issues/287

Description of changes:
- fix for this - https://github.com/awslabs/kubernetes-iteration-toolkit/issues/287
- also fix example tekton tasks that are obsolete w.r.t current state of KIT repo.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,13 @@
+# Kubernetes Iteration Toolkit Tests with Tekton
+
+### Overview:
+Tekton is a powerful and flexible open-source framework for creating CI/CD systems, allowing developers to build, test, and deploy across cloud providers and on-premise systems.
+
+
+
+### Test Images
+
+Tekton tasks can leverage image like clusterloader2 to perform various kinds of tests like load, pod-density on K8s cluster on KIT infra. 
+To build the docker image for clusterloader2 use the below command which takes `branch` as a build arg which lets us build clusterloader2 for a given branch on this repo[here](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2)
+
+- docker build --build-arg branch=release-1.23 ./images/clusterloader2/
--- a/tests/images/clusterloader2/Dockerfile
+++ b/tests/images/clusterloader2/Dockerfile
@@ -1,6 +1,9 @@
 FROM golang:1.16.4 AS builder
 WORKDIR /go/src/k8s.io
 RUN git clone https://github.com/kubernetes/perf-tests
+WORKDIR perf-tests
+RUN git fetch origin --verbose --tags
+RUN git checkout $branch
 WORKDIR /go/src/k8s.io/perf-tests/clusterloader2
 RUN GOPROXY=direct GOOS=linux CGO_ENABLED=0  go build -o ./clusterloader ./cmd
 
--- a/tests/tasks/generators/clusterloader/load.yaml
+++ b/tests/tasks/generators/clusterloader/load.yaml
@@ -10,6 +10,9 @@ spec:
   - name: giturl
     description: "git url to clone the package"
     default: https://github.com/kubernetes/perf-tests.git
+  - name: cl2-branch
+    description: "The branch of clusterloader2 you want to use"
+    default: "release-1.23"
   - name: nodes-per-namespace
     description: "nodes per namespace to get created for load test "
     default: "100"
@@ -39,7 +42,12 @@ spec:
   - name: git-clone      
     image: alpine/git
     workingDir: $(workspaces.source.path)
-    args: ["clone", "$(params.giturl)"]
+    script: |
+      git clone $(params.giturl)
+      cd $(workspaces.source.path)/perf-tests/
+      git fetch origin --verbose --tags
+      git checkout $(params.cl2-branch)
+      git branch
   - name: prepare-loadtest
     image: alpine/k8s:1.22.6
     workingDir: $(workspaces.source.path)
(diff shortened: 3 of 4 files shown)


tests/pipelines/kitctl/template.yaml
```diff
@@ -16,17 +16,10 @@ spec:
     params:
       - name: name
         value: '$(params.name)'
-  - name: setup-data-plane
```
prateekgogia, line 19: Why are we getting rid of dataplane setup from the template?

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -60,6 +63,10 @@ spec:
       NODE_MODE: master 
       EOL
       cat $(workspaces.source.path)/overrides.yaml
+      cd $(workspaces.source.path)/perf-tests/
+      git fetch origin --verbose --tags
+      git checkout $(params.cl2-branch)
+      git branch
```
prateekgogia, line 69: We should be able to checkout the tag in the earlier step `git-clone`
https://github.com/tektoncd/catalog/blob/main/task/git-clone/0.8/samples/using-git-clone-result.yaml#L36

tests/tasks/setup/kitctl/controlplane.yaml
```diff
@@ -68,8 +62,6 @@ spec:
         etcd:
           replicas: $(params.etcd-replicas)
           spec:
-            nodeSelector:
-              node.kubernetes.io/instance-type: $(params.etcd-instance-type)
```
prateekgogia, line 72: If we don't set this, how do make sure which instance the controlplane components will be running?

tests/tasks/setup/kitctl/dataplane.yaml
```diff
@@ -37,8 +35,6 @@ spec:
       spec:
         clusterName: $(params.name) # Associated Cluster name
         nodeCount: $(params.node-count)
-        subnetSelector:
-          karpenter.sh/discovery: ${TAG}
```
prateekgogia, line 41: Same question how does subnet get assigned?

tests/images/clusterloader2/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM golang:1.16.4 AS builder
 WORKDIR /go/src/k8s.io
 RUN git clone https://github.com/kubernetes/perf-tests
+WORKDIR perf-tests
+RUN git fetch origin --verbose --tags
+RUN git checkout $branch
```
prateekgogia, line 6: How does the $branch variable get set? Do you set it on your terminal before running docker build? Don't we need an `ARG`?
https://stackoverflow.com/a/51770027/3175021

hakuna-matatah: `docker build --build-arg branch=release-1.23 .` - this is how it is run. It's a docker feature to be able to pass build args.

tests/pipelines/kitctl/template.yaml
hakuna-matatah: This is just an example pipeline not really used for anything but demo. Current dataplane setup is obsolete and prevents this pipeline from succeeding.

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: Yeah, moving these commands to that step will make us to move to `script` approach to look clean. As it will be multiple commands which is why i added it here. I don't have hard preference on this.

tests/tasks/setup/kitctl/controlplane.yaml
hakuna-matatah: Setting this here like this will make it run likely into karpenter unable to provision anything at all when there is icing. Instead there is a way to set this up through Karpenter-kit provisioner (or) you can use different bundle-types etc variables that karp can understand.

tests/images/clusterloader2/Dockerfile
prateekgogia: Can we document this somewhere in the repo in the docs?

tests/pipelines/kitctl/template.yaml
prateekgogia: >Current dataplane setup is obsolete 

Why is it obsolete, whats the issue with the pipeline? If you don't set node count in the spec it won't provision any nodes.

tests/tasks/generators/clusterloader/load.yaml
prateekgogia: I think we should keep the git cloning and checkout logic in one place

tests/tasks/setup/kitctl/controlplane.yaml
prateekgogia: But we don't change the Karpenter provisioner in our tests. If I need to create a guest cluster how do I define the instance type if we remove this?

hakuna-matatah: But you'd run into issues by hardcoding it to instance type as i did if you hardcode instance types like this. Ideally you'd want to take this approach - https://karpenter.sh/preview/tasks/scheduling/#supported-labels

Anyways all these changes are clean up and can be done as part of other PR and don't want to muddy the water from the original issue the PR is raised for. Because these changes were already committed to this branch many days back which is why these ended up in this PR. We can come back to these later may be like a separate git issue for validating the current kit pipelines w.rt current state of KIT repo.

tests/README.md
```diff
@@ -0,0 +1,12 @@
...
+### Overview:
+Tekton is a powerful and flexible open-source framework for creating CI/CD systems, allowing developers to build, test, and deploy across cloud providers and on-premise systems.
+
+
+
+### Test Images
+
+Tekton tasks can leverage image like clusterloader2 to perform various kinds of tests like load, pod-density on K8s cluster on KIT infra. 
+To build the docker image for clusterloader2 use the below command which takes `branch` as a build arg which lets us build clusterloader2 for a given branch on this repo[here](https://github.com/kubernetes/perf-tests/tree/master/clusterloader2)
+- docker build --build-arg branch=release-1.23 .
```
prateekgogia, line 12: This command won't work in tests directory since Dockerfile is in [images/clusterloader2](https://github.com/awslabs/kubernetes-iteration-toolkit/tree/main/tests/images/clusterloader2)

prateekgogia (approved):

PR #293, opened 2022-09-13
Add codegen script & remove git diff check in `substrate` makefile

xdu31:
Issue #, if available:

Description of changes:
1. Add codegen script
2. remove git diff check in `substrate`
3. add change in `zz_generated.deepcopy.go`

### DIFF
--- a/substrate/Makefile
+++ b/substrate/Makefile
@@ -12,14 +12,8 @@ verify: codegen ## Verify code. Includes dependencies, linting, formatting, etc
 	go vet ./...
 	go fmt ./...
 	golangci-lint run
-	@git diff --quiet ||\
-		{ echo "New file modification detected in the Git working tree. Please check in before commit.";\
-		exit 1;}
 
 codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
-	controller-gen \
-		object:headerFile="hack/boilerplate.go.txt" \
-		paths="./pkg/..."
-	hack/boilerplate.sh
+	hack/codegen.sh
 
 .PHONY: help dev test verify
--- a/substrate/hack/codegen.sh
+++ b/substrate/hack/codegen.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+controller-gen \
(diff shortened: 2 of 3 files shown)


substrate/Makefile
```diff
@@ -12,14 +12,8 @@ verify: codegen ## Verify code. Includes dependencies, linting, formatting, etc
 	go vet ./...
 	go fmt ./...
 	golangci-lint run
-	@git diff --quiet ||\
-		{ echo "New file modification detected in the Git working tree. Please check in before commit.";\
-		exit 1;}
 
 codegen: ## Generate code. Must be run if changes are made to ./pkg/apis/...
-	controller-gen \
```
prateekgogia, line 20: whats the benefit for adding a new script?

substrate/Makefile
```diff
@@ -12,14 +12,8 @@ verify: codegen ## Verify code. Includes dependencies, linting, formatting, etc
 	go vet ./...
 	go fmt ./...
 	golangci-lint run
-	@git diff --quiet ||\
```
prateekgogia, line 15: Why are we removing this, does it not work as intended?

xdu31: > Why are we removing this, does it not work as intended?

Why do we have such a check on `git diff` in `substrate` but not in `operator`.
`make verify` is a general verification of the code, should be happening before code commit.
Additional reason, if we want to make a script with `make verify`, code commit is always a prerequisite because the `@git diff --quiet` will fail.

prateekgogia: maybe we can split this into different commands like
```
vendor:
     go mod tidy
     go mod download
fmt:
    go vet ./...
    go fmt ./...
verify: codegen vendor fmt
   golangci-lint run
   @git diff --quiet ....
```

The reason we want git diff is because we want to make sure in CI that there are no files someone has missed to update.

hakuna-matatah:
wondering if this is being worked on ? or need to be closed ?

hakuna-matatah:
Closing this PR. Please feel free to re-open when you want to update the changes.

PR #305, opened 2022-10-20
add a task to generate a timestamp

mengqiy:
It can be used as part of the `results-bucket` path to satisfy the build-id requirement of perfdash.

### DIFF
--- a/tests/pipelines/eks/awscli-eks-cl2-load.yaml
+++ b/tests/pipelines/eks/awscli-eks-cl2-load.yaml
@@ -21,6 +21,9 @@ spec:
   - name: slack-message
   - name: vpc-cfn-url
   tasks:
+  - name: timestamp
+    taskRef:
+      name: timestamp
   - name: slack-notification
     params:
     - name: slack-hook
--- a/tests/tasks/timestamp/timestamp.yaml
+++ b/tests/tasks/timestamp/timestamp.yaml
@@ -1,0 +1,15 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: timestamp
+  namespace: tekton-pipelines
+spec:
+  results:
+    - name: timestamp
+      description: The current date in unix timestamp format
+  steps:
+    - name: date
+      image: bash:latest
+      script: |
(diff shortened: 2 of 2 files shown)


tests/pipelines/eks/awscli-eks-cl2-load.yaml
```diff
@@ -21,6 +21,9 @@ spec:
   - name: slack-message
   - name: vpc-cfn-url
   tasks:
+  - name: timestamp
```
hakuna-matatah, line 24: do we need timestamp as a separate task ?
wdyt about adding same timestamp as part of pushing the results here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tasks/generators/clusterloader/load.yaml#L91 ?

mengqiy: I'm trying to use the following in the `TriggerTemplate`s:
```
- name: results-bucket
  value: kit-eks-scalability/kit-eks-1k/$(tasks.timestamp.results.timestamp)
```

Your approach works and simplify the wiring between tasks and triggertemplate.
But it changes the meaning of parameter `results-bucket`. It becomes to mean `results-bucket-prefix`. We will need to change the parameter name across the board.
I don't have a strong opinion here.

mengqiy: Had an offline discussion with Harish. The reusability is not important for this use case, since task `load` is the only one relies on it. We are not going to create a separate task just for generating the timestamp.
I will push the new change shortly.

hakuna-matatah: > We have added the timestamp when passing the parameters in tekton `TriggerTemplate` hosted in the internal repo. We no longer need this PR. Closing



> We have added the timestamp when passing the parameters in tekton `TriggerTemplate` hosted in the internal repo. We no longer need this PR. Closing

Thanks David.

hakuna-matatah:
> We have added the timestamp when passing the parameters in tekton `TriggerTemplate` hosted in the internal repo. We no longer need this PR. Closing

Thanks David.

PR #306, opened 2022-10-21
add perfdash in CDK

mengqiy:
Added perfdash components (excluding configmap) in CDK.
Added a section in the README for perfdash.

### DIFF
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -8,7 +8,7 @@ All other components are able to be installed via Flux resources.
 
 ### Components:
 
-KIT Infrastructure creates a base K8s cluster with below  add-ons by default  but also provides an ability to not install some of these addons optionally through CDK context. Add-ons include permissions scoped to the pod using IAM Roles for Service Accounts (IRSA).
+KIT Infrastructure creates a base K8s cluster with below add-ons by default but also provides an ability to not install some of these addons optionally through CDK context. Add-ons include permissions scoped to the pod using IAM Roles for Service Accounts (IRSA).
 
 - EKS Cluster (host cluster)
 - EKS Managed Node Group (for critical add-ons mentioned below)
@@ -18,7 +18,53 @@ KIT Infrastructure creates a base K8s cluster with below  add-ons by default  bu
 - Flux v2
 - Kubernetes Iteration Toolkit (KIT) Operator (optional)
 
-Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, and the metrics-server. 
+Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, the metrics-server and [perf dash].
+
+#### Perf Dash
+
+Perf dash is an optional addon and disabled by default.
+
+To bring it up, there are a few places you need to ensure they are wired correctly.
+
+- A k8s service account named `perfdash-log-fetcher` is expected and has the permissions to fetch the logs from the s3 bucket.
+- A ConfigMap containing the target S3 bucket name that hosts the log and the jobs config for perf dash must be created so that it can be mounted as volume in the Deployment. An example configmap can be:
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: perfdash-config
+  namespace: perfdash
+  labels:
+    app.kubernetes.io/name: perfdash
+    app.kubernetes.io/component: config
+    app.kubernetes.io/part-of: perfdash
+data:
+  PERFDASH_LOG_BUCKET: my-bucket-name
+  jobs.yaml: |
+    periodics:
+      - name: "kit-eks-1k"
+        tags:
+          - "perfDashPrefix: 1K test"
+          - "perfDashJobType: performance"
+      - name: "kit-eks-2k"
+        tags:
+        - "perfDashPrefix: 2k test"
+        - "perfDashJobType: performance"
+```
+- The logs for different runs in the bucket should have the layout like the following:
+```
+my-s3-bucket
+├── kit-eks-1k # note: this should match the name in perf dash config.
+│   ├── 123
+│   │   └── cl2-logs
+│   └── 124
+│       └── cl2-logs
+└── kit-eks-2k # note: this should match the name in perf dash config.
+    ├── 125
+    │   └── cl2-logs
+    └── 126
+        └── cl2-logs
+```
 
 ### Getting Started:
 
@@ -30,7 +76,7 @@ To launch the KIT infrastructure, ensure you have the following installed:
 By default, the CDK application will wire up a Flux config that will monitor for k8s manifests within the KIT repo at `infrastructure/k8s-config/clusters/kit-infrastructure`.
 The parameters supplied to the CDK app will allow you to wire up your own application's repo so that you can place tekton test files and other cluster components there.
 
- As an example, below are the parameters used for the [Karpenter](https://github.com/aws/karpenter) project.
+As an example, below are the parameters used for the [Karpenter](https://github.com/aws/karpenter) project.
 
  ```shell
 cdk bootstrap
@@ -43,8 +89,9 @@ cdk deploy KITInfrastructure --no-rollback \
   -c TestServiceAccount="karpenter-tests"
  ```
 
-  As an example, below are the parmeters used if you want to selectively disable some addons like Karpenter, EBSCSIDriver, KIT.
- ```shell
+As an example, below are the parmeters used if you want to selectively enable some addons like perfdash and disable some addons like Karpenter, EBSCSIDriver, KIT.
+
+```shell
 cdk bootstrap
 cdk deploy KITInfrastructure --no-rollback \
   -c TestNamespace="tekton-pipelines" \
@@ -52,7 +99,8 @@ cdk deploy KITInfrastructure --no-rollback \
   -c AWSEBSCSIDriverAddon=false \
   -c KarpenterAddon=false \
   -c KITAddon=false \
- ```
+  -c FluxAddonPaths="./test/infrastructure/clusters/addons/perfdash"
+```
 
 ### Context Parameters:
 
@@ -61,6 +109,7 @@ cdk deploy KITInfrastructure --no-rollback \
 | FluxRepoURL         | Flux Source git repo URL to synchronize KIT infrastructure like Tekton                     | https://github.com/awslabs/kubernetes-iteration-toolkit |   |   |
 | FluxRepoBranch      | Flux Source git repo branch to synchronize KIT infrastructure                              | main                                                    |   |   |
 | FluxRepoPath        | Flux Source git repo path to Kubernetes resources                                          | ./infrastructure/k8s-config/clusters/kit-infrastructure |   |   |
+| FluxAddonPaths      | Flux Source git repo paths (separted by comma) to Kubernetes addons.                       |                                                         |   |   |
 | TestFluxRepoName    | Flux Source git repo name to synchronize application tests like Tekton Tasks and Pipelines |                                                         |   |   |
 | TestFluxRepoURL     | Flux Source git repo URL to synchronize application tests                                  |                                                         |   |   |
 | TestFluxRepoBranch  | Flux Source git repo branch to synchronize application tests                               |                                                         |   |   |
@@ -70,3 +119,5 @@ cdk deploy KITInfrastructure --no-rollback \
 | KITAddon            | KIT CRD addon that gets installed on KIT Infrastructure by default                         | true                                                    |   |   |                               
 | KarpenterAddon      | Karpenter CRD addon that gets installed on KIT Infrastructure by default                   | true                                                    |   |   | 
 | AWSEBSCSIDriverAddon| AWSEBSCSIDriver CRD addon that gets installed on KIT Infrastructure by default             | true                                                    |   |   |
+
+[perf dash]: https://github.com/kubernetes/perf-tests/tree/master/perfdash
--- a/infrastructure/lib/addons/fluxv2.ts
+++ b/infrastructure/lib/addons/fluxv2.ts
@@ -10,6 +10,7 @@ export interface FluxV2Props extends StackProps {
   repoUrl: string;
   repoBranch: string;
   repoPath: string;
+  repoAddonPaths?: string;
   testRepoName: string;
   testRepoUrl?: string;
   testRepoBranch?: string;
@@ -87,7 +88,7 @@ export class FluxV2 extends Construct {
             ],
           },
       }
-  });
+    });
 
   // Bootstrap manifests
   const gitRepoManifest = props.cluster.addManifest('GitRepoSelf', {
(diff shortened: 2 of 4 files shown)


infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
```diff
@@ -0,0 +1,118 @@
...
+kind: ConfigMap
+metadata:
+  name: perfdash-config
+  namespace: perfdash
+  labels:
+    app.kubernetes.io/name: perfdash
+    app.kubernetes.io/component: config
+    app.kubernetes.io/part-of: perfdash
+data:
+  jobs.yaml: |
```
hakuna-matatah, line 16: - This data about jobs are more specific to internal things that we run. It doesn't make sense to keep this in upstream repo. Could we push this in our internal repo.
wdyt ?

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
```diff
@@ -0,0 +1,118 @@
...
+        app.kubernetes.io/part-of: perfdash
+    spec:
+      containers:
+        - name: perfdash
+          image: gcr.io/k8s-staging-perf-tests/perfdash:2.44
+          command:
+            - /perfdash
+            - --www=true
+            - --address=0.0.0.0:8080
+            - --builds=20
```
hakuna-matatah, line 75: IIUC does this mean we only display 20 builds on dashboard ?

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
```diff
@@ -0,0 +1,118 @@
...
+            - --force-builds
+            - --mode=s3
+            - --aws-region=us-west-2
+            - --configPath=/etc/config/jobs.yaml
+            - --logsBucket=kit-eks-scalability
```
hakuna-matatah, line 80: Either we need to make sure we explicitly document this detail because other customers outside of us won't have access to this bucket and this deployment pods might fail or throw errors in the logs (OR) move this to internal repo ? Or patch this specific detail through flux  by taking in as param through cdk command line args ?

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
```diff
@@ -0,0 +1,118 @@
...
+            - --logsPath=
```
hakuna-matatah, line 81: any reason for having this here with empty value ?

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
mengqiy: That makes sense.

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
mengqiy: What's your suggestion? 50 or 100 or some other value?

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
mengqiy: It appears upstream has another layer called `logs` in the hierarchy. We don't have that and it's not necessary for our use case. So we have it as an empty value.

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
mengqiy: I don't quite understand this comment.
Who is other customers? Other EKS teams or EKS customers?
IIUC this is bucket is only accessible to our team. Others can't access it no matter where we keep this yaml file.
Can you please elaborate your concern? We can discuss offline if you want.

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
hakuna-matatah: cool, i see it here - https://github.com/kubernetes/perf-tests/blob/master/perfdash/perfdash.go#L52

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
mengqiy: Updated according to our offline discussion.

mengqiy:
PTAL

mengqiy:
@hakuna-matatah Please note that I have created an internal CR for the ConfigMap. We need to merge and apply that before we merge this PR.

infrastructure/README.md
```diff
@@ -18,7 +18,52 @@ KIT Infrastructure creates a base K8s cluster with below  add-ons by default  bu
...
+Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, the metrics-server and [perf dash].
+
+#### Perf Dash
+
+Perf dash is optional. If you want to disable it, you can simply remove or comment out the content in `./infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml`.
+
+To bring it up, there are a few places you need to ensure they are wired correctly.
+
+- The `logsBucket` logs bucket should be your bucket holding the logs. Ideally, it should be in the same AWS account as your KIT Infrastructure. Otherwise, you may need to configure the auth yourself.
+- A ConfigMap containing the jobs config for perf dash must be created so that it can be mounted as volumn in the Deployment. An example configmap can be:
```
hakuna-matatah, line 30: Nit: volum -> volume

infrastructure/README.md
```diff
@@ -18,7 +18,52 @@ KIT Infrastructure creates a base K8s cluster with below  add-ons by default  bu
...
-Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, and the metrics-server. 
+Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, the metrics-server and [perf dash].
+
+#### Perf Dash
+
+Perf dash is optional. If you want to disable it, you can simply remove or comment out the content in `./infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml`.
+
+To bring it up, there are a few places you need to ensure they are wired correctly.
+
+- The `logsBucket` logs bucket should be your bucket holding the logs. Ideally, it should be in the same AWS account as your KIT Infrastructure. Otherwise, you may need to configure the auth yourself.
```
hakuna-matatah, line 29: Thanks for adding these details.

infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml
```diff
@@ -0,0 +1,81 @@
...
+        app.kubernetes.io/part-of: perfdash
+    spec:
+      containers:
+        - name: perfdash
+          image: gcr.io/k8s-staging-perf-tests/perfdash:2.44
+          command:
+            - /perfdash
+            - --www=true
+            - --address=0.0.0.0:8080
+            - --builds=100
```
hakuna-matatah, line 38: Nice!

hakuna-matatah:
> @hakuna-matatah Please note that I have created an internal CR for the ConfigMap. We need to merge and apply that before we merge this PR.
> 
> 

Yup, thanks for moving that to internal.

infrastructure/README.md
mengqiy: Done

mengqiy:
@hakuna-matatah  PTAL

hakuna-matatah (dismissed):
lgtm

mengqiy:
This PR should hold merging until the ConfigMap has been created out-of-the-band.

infrastructure/README.md
```diff
@@ -18,7 +18,52 @@ KIT Infrastructure creates a base K8s cluster with below  add-ons by default  bu
 - Flux v2
 - Kubernetes Iteration Toolkit (KIT) Operator (optional)
 
-Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, and the metrics-server. 
+Flux is setup, by deafult, to monitor the KIT git repo path `./infrastructure/k8s-config/clusters/kit-infrastructure`, which includes other add-ons that do not require AWS credentials such as tekton, prometheus, grafana, the metrics-server and [perf dash].
```
prateekgogia, line 21: Curious why do we have perf dash text formatting like this - `[perf dash]`?

infrastructure/README.md
```diff
@@ -18,7 +18,52 @@ KIT Infrastructure creates a base K8s cluster with below  add-ons by default  bu
...
+
+#### Perf Dash
+
+Perf dash is optional. If you want to disable it, you can simply remove or comment out the content in `./infrastructure/k8s-config/clusters/kit-infrastructure/perfdash/perfdash.yaml`.
```
prateekgogia, line 25: Doesn't provide info if its enabled by default? 
> If you want to disable it, you can simply remove or comment out the content in

I don't think this will work because it will be synced from Flux and we don't want everyone to be changing KIT repo for disabling perfdash. 

Also enabling by default would mean existing clusters in Karpenter will get this installed which might be a surprise for them. I would minimize such changes in this repo or at least get a consent from Karpenter team.
cc @bwagner5

infrastructure/README.md
```diff
@@ -69,4 +115,6 @@ cdk deploy KITInfrastructure --no-rollback \
 | TestServiceAccount  | Service Account for application tests to run with                                          |                                                         |   |   |
 | KITAddon            | KIT CRD addon that gets installed on KIT Infrastructure by default                         | true                                                    |   |   |                               
 | KarpenterAddon      | Karpenter CRD addon that gets installed on KIT Infrastructure by default                   | true                                                    |   |   | 
-| AWSEBSCSIDriverAddon| AWSEBSCSIDriver CRD addon that gets installed on KIT Infrastructure by default             | true                                                    |   |   |
\ No newline at end of file
+| AWSEBSCSIDriverAddon| AWSEBSCSIDriver CRD addon that gets installed on KIT Infrastructure by default             | true                                                    |   |   |
+
+[perf dash]: https://github.com/kubernetes/perf-tests/tree/master/perfdash
```
prateekgogia, line 120: Why different formatting?

prateekgogia (dismissed):

infrastructure/README.md
mengqiy: It makes the raw markdown easier to read when the long URLs is not embedded in the text.
The link of `perf dash` is defined at the bottom of this file.

infrastructure/README.md
mengqiy: Thanks for bringing this up! It seems karpenter do use KIT for testing: https://github.com/aws/karpenter/tree/main/test/infrastructure#setup.

>I don't think this will work because it will be synced from Flux and we don't want everyone to be changing KIT repo for disabling perfdash.

I thought a user cloud have their own forks where they can make some changes. I can rephrase it to make it more clear. If you don't think that's a valid assumption, I can change it.

>Also enabling by default would mean existing clusters in Karpenter will get this installed which might be a surprise for them.

I guess it won't be a surprise for Karpenter team, since it doesn't break their cluster. But I guess this addon is not useful for them, since they don't CL2 tests at all. We may not want to install it default.
An alternative is to add it as an addon under `./infrastructure/lib/addons/` using typescript. A flag can be used to gate it and make it disabled by default.
@prateekgogia @hakuna-matatah WDYT?

infrastructure/README.md
mengqiy: Answered in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/306#discussion_r1009719638

infrastructure/README.md
hakuna-matatah: Current state of things - We install certain addons today through typescript cdk thing if that addon requires aws resources to function otherwise we install them through flux. Customers have an option of disabling certain addons through cdk if they are being installed but not through flux.
We need to take a stand on whether we want to move all the addons to be installed through type-script and give customers an option of disabling unwanted addons (OR) have a separate directory or git repo to be pointed out for all optional addons so customers have flexibility to choose at the time of setup. For the latter one (i.e; flux route), all we have to do is basically send an array list here for repoPaths and repos, so customers can pass it at the time of setup  - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L167-L177

mengqiy: > have a separate directory or git repo to be pointed out for all optional addons so customers have flexibility to choose at the time of setup.

I guess this approach makes more sense.

mengqiy:
Rebased on top of the latest HEAD. The last commit is the change since last review.
The changes is not fully tested due to https://github.com/awslabs/kubernetes-iteration-toolkit/issues/313. I will need some help to figure out what's going on with #313.

infrastructure/lib/addons/fluxv2.ts
```diff
@@ -10,10 +10,12 @@ export interface FluxV2Props extends StackProps {
   repoUrl: string;
   repoBranch: string;
   repoPath: string;
+  repoAddonPaths?: string[];
   testRepoName: string;
   testRepoUrl?: string;
   testRepoBranch?: string;
   testRepoPath?: string;
+  testRepoAddonPaths?: string[];
```
hakuna-matatah, line 18: why would you  need two arrays like one [here](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/306/files#diff-bbf436cf80554d1c16a136a08cbc43d87d488d7ab9d51b6ca1aa66a74de555b9R13) and on this line ? Can't we just have one.

infrastructure/lib/addons/fluxv2.ts
```diff
@@ -127,6 +129,30 @@ export class FluxV2 extends Construct {
   });
   kustomizationManifest.node.addDependency(chart);
 
+  if (props.repoAddonPaths) {
+    for (var repoAddonPath of props.repoAddonPaths) {
+      const addonKustomizationManifest = props.cluster.addManifest('AddonKustomizationSelf', {
```
hakuna-matatah, line 134: could we wrap this in `function` as we are using it multiple times ?

infrastructure/lib/addons/fluxv2.ts
mengqiy: From my current understanding, we support `Repo` and `testRepo` (this is not a great name, but we are using it now), user may want to install the addons from base `Repo` and (or) `testRepo`. We should provide both options IMO.

infrastructure/lib/addons/fluxv2.ts
mengqiy: That's possible.
https://github.com/awslabs/kubernetes-iteration-toolkit/blob/da36db42fe000ace00c9396fb3abd0af226da3b0/infrastructure/lib/addons/fluxv2.ts#L110-L128
and https://github.com/awslabs/kubernetes-iteration-toolkit/blob/da36db42fe000ace00c9396fb3abd0af226da3b0/infrastructure/lib/addons/fluxv2.ts#L150-L168 are doing similar things for `Repo` and `testRepo`. We should clean them up as well eventually.

infrastructure/lib/addons/fluxv2.ts
hakuna-matatah: The idea there was to have `repo` for k8s addons that needs to be installed on Management cluster through Flux and `testRepo` is a place holder for tekton tasks tests etc that we wanted to separate out at that time from the Infrastructure code repository itself. 


Given with this PR, now we can take in array of repos, this logical separation of `repos` and `testRepos` may not be needed at this point. Because customers can bring in any number of arbitrary set of `testrepos` which contain tests to sync on cluster and/or also arbitrary set of k8s addons can be installed  `repos` using the same array list . Doesn't add any benefit now to have separate vars for this. Makes sense ?

mengqiy:
@hakuna-matatah PTAL

hakuna-matatah (approved):
lgtm

PR #307, opened 2022-10-24
Use the kubectl image that contains curl

mengqiy:
tests/tasks/setup/kitctl/controlplane.yaml uses curl to probe the healthz endpoint.
It's not trivial to install curl when we are already running it as user "nobody".

### DIFF
--- a/tests/tasks/setup/kitctl/controlplane.yaml
+++ b/tests/tasks/setup/kitctl/controlplane.yaml
@@ -46,7 +46,7 @@ spec:
     description: Kubernetes version for the guest cluster 
   steps:
   - name: setup-control-plane
-    image: bitnami/kubectl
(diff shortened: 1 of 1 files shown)


tests/tasks/setup/kitctl/controlplane.yaml
```diff
@@ -46,7 +46,7 @@ spec:
     description: Kubernetes version for the guest cluster 
   steps:
   - name: setup-control-plane
-    image: bitnami/kubectl
+    image: bitnami/kubectl:1.24.5 # curl was removed in more recent versions
```
prateekgogia, line 49: Fine for now but we will need to find a better way once 1.24 is EOL.

prateekgogia (approved):

PR #311, opened 2022-11-02
add ci for CDK

mengqiy:
Add a basic CI (build only since we don't have any tests) for CDK.


hakuna-matatah (approved):
Thanks for adding this.

LGTM!

PR #314, opened 2022-11-03
Use resource-name/object-name in kubectl set env

mengqiy:
Perf https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-env-em-


hakuna-matatah (approved):
lgtm

PR #316, opened 2022-11-04
Tekton task to emit cloudwatch metrics for Cl2 tests

Ganiredi:
Issue #, if available:

Description of changes:
This task emits datapoint for cl2 tests to cloud watch ,if the cl2 test passes it will emit 0 as data point otherwise its emit data point as 1.

### DIFF
--- a/tests/tasks/Cloudwatch/cw-metric.yaml
+++ b/tests/tasks/Cloudwatch/cw-metric.yaml
@@ -4,7 +4,7 @@ metadata:
   name:  cloudwatch
   namespace: tekton-pipelines
 spec:
-  description: "cw task to push metrics relying on generate task."
+  description: "cw task to push one data point to AWS CloudWatch."
   params:
     - name: region
       default: "us-west-2"
@@ -12,10 +12,11 @@ spec:
     - name: load-result
   steps:
   - name: cw-emit
-    image: alpine/k8s:1.22.6
+    image: amazon/aws-cli
     script: |
+      aws sts get-caller-identity
       if [[ "$(params.load-result)" -eq 0 ]]; then
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name cl2-loadtest --namespace k8s-version --unit Count --value 1 --dimensions Nodes=$(params.desired-nodes)
       else
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 0 --dimensions TypeOfTest=Cl2_Test 
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name cl2-loadtest --namespace k8s-version --unit Count --value 0 --dimensions Nodes=$(params.desired-nodes)
       fi
--- a/tests/tasks/Cloudwatch/cw-metric.yaml
+++ b/tests/tasks/Cloudwatch/cw-metric.yaml
@@ -4,18 +4,26 @@ metadata:
   name:  cloudwatch
   namespace: tekton-pipelines
 spec:
-  description: "cw task to push metrics relying on generate task."
+  description: "cw task that can be used to create a metric and emit a datapoint based on the result that is passed after test."
   params:
     - name: region
       default: "us-west-2"
-    - name: desired-nodes
-    - name: load-result
+      description: "The region to use for publishing the metrics"
+    - name: metric-name 
+      default: "cl2-loadtest"
+      description: "The name of the metric you want to pass"
+    - name: namespace
+      description: "The namespace for the metric data"
+    - name: dimensions
+      description: "Name of the dimension that you want to associate with the metric"
+    - name: unit
+      default: "Count"
+      description: "Name of the unit you want to use when storing the metric."
+    - name: value
+      description: "The value for the metric"
   steps:
   - name: cw-emit
-    image: alpine/k8s:1.22.6
+    image: amazon/aws-cli
     script: |
-      if [[ "$(params.load-result)" -eq 0 ]]; then
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
-      else
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 0 --dimensions TypeOfTest=Cl2_Test 
-      fi
+      aws sts get-caller-identity
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.metric-name) --namespace $(params.namespace) --dimensions Nodes=$(params.dimensions) --unit $(params.unit) --value $(params.value) 
--- a/tests/tasks/Cloudwatch/cw-metric.yaml
+++ b/tests/tasks/Cloudwatch/cw-metric.yaml
@@ -1,0 +1,22 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name:  cloudwatch
+  namespace: tekton-pipelines
+spec:
+  description: "cw task to push metrics relying on generate task."
+  params:
+    - name: region
+      default: "us-west-2"
+    - name: desired-nodes
+    - name: load-result
+  steps:
+  - name: cw-emit
+    image: alpine/k8s:1.22.6
+    script: |
+      #!/bin/bash
+      if [[ "$(params.load-result)" -eq 0 ]]; then
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name Cl2Test --namespace Cl2Load-Tests --unit Count --value 0 --dimensions NumberofNodes=$(params.desired-nodes) 
+      else
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name Cl2Test --namespace Cl2Load-Tests --unit Count --value 1 --dimensions NumberofNodes=$(params.desired-nodes) 
+      fi
--- a/tests/tasks/Cloudwatch/cw-metric.yaml
+++ b/tests/tasks/Cloudwatch/cw-metric.yaml
@@ -14,9 +14,8 @@ spec:
   - name: cw-emit
     image: alpine/k8s:1.22.6
     script: |
-      #!/bin/bash
       if [[ "$(params.load-result)" -eq 0 ]]; then
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name Cl2Test --namespace Cl2Load-Tests --unit Count --value 0 --dimensions NumberofNodes=$(params.desired-nodes) 
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
       else
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name Cl2Test --namespace Cl2Load-Tests --unit Count --value 1 --dimensions NumberofNodes=$(params.desired-nodes) 
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 0 --dimensions TypeOfTest=Cl2_Test 
       fi
--- a/tests/tasks/Cloudwatch/cw-metric.yaml
+++ b/tests/tasks/Cloudwatch/cw-metric.yaml
@@ -4,7 +4,7 @@ metadata:
   name:  cloudwatch
   namespace: tekton-pipelines
 spec:
-  description: "cw task that can be used to create a metric and emit a datapoint based on the result that is passed after test."
+  description: "Cloudwatch task publishes the provided metric data points to Amazon CloudWatch. More details can be found in the aws cli reference doc: https://awscli.amazonaws.com/v2/documentation/api/2.1.29/reference/cloudwatch/put-metric-data.html"
   params:
     - name: region
       default: "us-west-2"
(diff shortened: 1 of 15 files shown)


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -35,6 +35,9 @@ spec:
     description: "aws eks enpoint to create clusters against"
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  results:
+    - name: datapoint
+      description: Stores the cl2 loadtest result to emit cw metrics to aws cloudwatch
```
mengqiy, line 40: description: Stores the cl2 loadtest result to be consumed by others (e.g. emitting to CloudWatch)

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -80,6 +83,7 @@ spec:
       aws eks $ENDPOINT_FLAG update-kubeconfig --name $(params.cluster-name) --region $(params.region)
```
mengqiy, line 83: I guess you need to add `#!/bin/bash` at the beginning of the `script` field, since we don't want `set -e` which is the default.
Ref: https://tekton.dev/docs/pipelines/tasks/#running-scripts-within-steps

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,22 @@
...
+  params:
+    - name: region
+      default: "us-west-2"
+    - name: desired-nodes
+    - name: load-result
+  steps:
+  - name: cw-emit
+    image: alpine/k8s:1.22.6
+    script: |
+      #!/bin/bash
```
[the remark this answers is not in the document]
Ganiredi: Are you mentioning about the script?

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  description: "cw task to push metrics relying on generate task."
+  params:
+    - name: region
+      default: "us-west-2"
+    - name: desired-nodes
+    - name: load-result
+  steps:
+  - name: cw-emit
+    image: alpine/k8s:1.22.6
```
mengqiy, line 15: We don't depend on any k8s stuffs in this task. Why don't we use image `amaon/aws-cli`?

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,21 @@
...
+    script: |
+      if [[ "$(params.load-result)" -eq 0 ]]; then
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
```
mengqiy, line 18: How does it know which AWS account to use?

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name:  cloudwatch
+  namespace: tekton-pipelines
+spec:
+  description: "cw task to push metrics relying on generate task."
```
mengqiy, line 7: We should talk about what it does in the description. e.g. `cw task push one data point to AWS CloudWatch`.

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: desired-nodes
+    - name: load-result
+  steps:
+  - name: cw-emit
+    image: alpine/k8s:1.22.6
+    script: |
+      if [[ "$(params.load-result)" -eq 0 ]]; then
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
+      else
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 0 --dimensions TypeOfTest=Cl2_Test
```
mengqiy, line 20: `$(params.desired-nodes)_nodes` should not be the`metric-name`. Number of nodes is a dimension. The `metric-name` can be something like `success`.
Also why do we want dimension of `TypeOfTest=Cl2_Test`, do we expect other type of tests?
@hakuna-matatah WDYT? We should get this right. It'd be harder to change it later.

Ganiredi: After performing some tests ,i come up with that name ,Will be changing accordingly as per your suggestions.

hakuna-matatah: We can have metric name as `cl2-loadtest` and dimension as nodes and namespace would be `k8s version`
In future we would be running different types of tests like density etc and we need to differentiate these types of test with `loadtests` we are running today.

tests/tasks/Cloudwatch/cw-metric.yaml
Ganiredi: I believe whatever the aws creds that we were using o run up these pipelines ,it will consider the same account ,unless if we want specify any.

tests/tasks/Cloudwatch/cw-metric.yaml
Ganiredi: yes correct will change it

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -4,18 +4,19 @@ metadata:
...
-    image: alpine/k8s:1.22.6
+    image: amazon/aws-cli
     script: |
+      aws sts get-caller-identity
       if [[ "$(params.load-result)" -eq 0 ]]; then
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 1 --dimensions TypeOfTest=Cl2_Test
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name cl2-loadtest --namespace k8s-version --unit Count --value 1 --dimensions Nodes=$(params.desired-nodes)
       else
-      aws cloudwatch --region $(params.region) put-metric-data --metric-name $(params.desired-nodes)_nodes --namespace Cl2Load-Tests --unit Count --value 0 --dimensions TypeOfTest=Cl2_Test 
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name cl2-loadtest --namespace k8s-version --unit Count --value 0 --dimensions Nodes=$(params.desired-nodes)
```
mengqiy, line 21: Per @hakuna-matatah's comment, we still want to keep dimensions `TypeOfTest`. We can use `TypeOfTest=load` and add `TypeOfTest=density` in the future.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -73,13 +76,15 @@ spec:
...
       if [ -n "$(params.endpoint)" ]; then
         ENDPOINT_FLAG="--endpoint $(params.endpoint)"
       fi
       aws eks $ENDPOINT_FLAG update-kubeconfig --name $(params.cluster-name) --region $(params.region)
       cat $(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml
       ENABLE_EXEC_SERVICE=false /clusterloader --kubeconfig=/root/.kube/config --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr --v=2
+      echo $? | tee $(results.cl2-result.path) 
     timeout: 30000s
   - name: upload-results
     image: alpine/k8s:1.22.6
```
mengqiy, line 90: @hakuna-matatah Do we have any best practice about which image to use between `alpine/k8s` and `amazon/aws-cli`? It seems `alpine/k8s` supports awscli v1 while `amazon/aws-cli` support awscli v2. Should we use awscli v2 here?

tests/tasks/Cloudwatch/cw-metric.yaml
hakuna-matatah: I don't think `type of test` dimension is required as we are explicitly having different metric name for different cl2 test type ( I suggested in my earlier comment like for example -  metric-name to be `cl2-loadtest`) . Makes sense ?

mengqiy: So `type of test` is in the metrics name instead of using a dimension. We will metrics like `cl2-loadtest` and `cl2-densitytest` (in the future).
It makes sense. It seems I misinterpreted your earlier comment.

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,22 @@
...
+      default: "us-west-2"
+    - name: desired-nodes
+    - name: load-result
+  steps:
+  - name: cw-emit
+    image: amazon/aws-cli
+    script: |
+      aws sts get-caller-identity
+      if [[ "$(params.load-result)" -eq 0 ]]; then
+      aws cloudwatch --region $(params.region) put-metric-data --metric-name cl2-loadtest --namespace k8s-version --unit Count --value 1 --dimensions Nodes=$(params.desired-nodes)
```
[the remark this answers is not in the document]
hakuna-matatah: The params for this tasks should be `dimensions`, `metricname`, `namespace` with meaningful defaults in our pipelines/task so it's much generic for anyone to use it outside of our use-case, after-all it is for emitting any CW metric. 

Also, having `K8s-version` as namespace would make this task tightly couple to always have `namespace` to `k8s version`.

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: Task
+metadata:
+  name:  cloudwatch
+  namespace: tekton-pipelines
+spec:
+  description: "cw task to push one data point to AWS CloudWatch."
+  params:
+    - name: region
+      default: "us-west-2"
+    - name: desired-nodes
```
hakuna-matatah, line 11: Please have generic naming so we don't tie this task to only emit metrics based on nodes. We can abstract it out and pass meaning values through our pipelines.

tests/tasks/Cloudwatch/cw-metric.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - name: load-result
```
hakuna-matatah, line 12: Same here

tests/tasks/Cloudwatch/cw-metric.yaml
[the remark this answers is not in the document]
mengqiy: It makes sense to make it generic.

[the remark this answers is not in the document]
mengqiy: Also to be generic enough that works for most use cases, --unit and --value should be parameterized (with reasonable default). It won't be very useful to others if the value is always 1.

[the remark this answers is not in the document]
mengqiy: Had a discussion with Ganesh offline. There is still some coupling: we need to convert `test-result` to `value`. This conversion has to happen somewhere. Either here or in the loadtest.
If we want to make this task generic enough, we may want the conversion to happen in the loadtest task.
@hakuna-matatah Thought?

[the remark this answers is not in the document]
hakuna-matatah: I agree with you David.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -35,6 +35,9 @@ spec:
     description: "aws eks enpoint to create clusters against"
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  results:
+    - name: datapoint
+      description: Stores the cl2 loadtest result to be consumed by cloudwatch
```
mengqiy, line 40: Consuming it is optional and any task besides CW can consume it. So I'd suggest the following wording:
`Stores the CL2 result that can be consumed by other tasks (e.g. cloudwatch)`

mengqiy (commented):
A few final nits. LGTM otherwise.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -35,6 +35,9 @@ spec:
     description: "aws eks enpoint to create clusters against"
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  results:
+    - name: datapoint
```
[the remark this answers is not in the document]
hakuna-matatah: can you take care of this comment @Ganiredi  ?

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -73,16 +76,21 @@ spec:
     image: public.ecr.aws/kit/clusterloader2:0213bea
     onError: continue
     script: |
+      #!/bin/bash
```
hakuna-matatah, line 79: why do you need this ?

Ganiredi: `fork/exec /tekton/scripts/script-0-424g4: exec format error` ,In my case i got this error ,But after i added the she bang command ,It resolved ,By default its set -e

tests/tasks/generators/clusterloader/load.yaml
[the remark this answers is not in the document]
hakuna-matatah: I think it's ok for now.

hakuna-matatah:
Thanks for the changes Ganesh. Could you test this change out in pipeline and post the output just to make sure we don't break our periodics once this is merged.

tests/tasks/generators/clusterloader/load.yaml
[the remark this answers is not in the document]
Ganiredi: okay @hakuna-matatah

tests/tasks/generators/clusterloader/load.yaml
mengqiy: I believe we need this here. If not adding shebang, tekton will prepend
```
#!/bin/bash
set -e
```
which will make the script exit when we get a non-zero exit code. It prevents us from collecting the CL2 exit code if it's not zero. Adding `#!/bin/bash` ourselves to ensure `set -e` is not added.

mengqiy (approved):
LGTM
We can merge it only after Ganesh can confirm it works in the manual testing.

mengqiy (approved):

Ganiredi:
image


image

image


```Run the test with two tasks as cw-task relies on load task ,Tests are running and attached the cloud output result

mengqiy:
I'm going to merge this now. I will kick off a test pipeline run to confirm everything works. If we encounter any issue, @Ganiredi may still have some time to fix it.

mengqiy:
There was an issue. We got it fixed in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/319

PR #325, opened 2022-11-30
Adds Remote Write changes for Prometheus

ashishranjan738:
This commit adds changes to remote write Prometheus metrics to a AMP workspace.

### DIFF
--- a/tests/tasks/generators/clusterloader/load.yaml
+++ b/tests/tasks/generators/clusterloader/load.yaml
@@ -35,11 +35,14 @@ spec:
     description: "aws eks enpoint to create clusters against"
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  - name: amp-workspace-id
+    description: The remote amp workspace id where remote needs to happen.
   results:
     - name: datapoint
       description: Stores the CL2 result that can be consumed by other tasks (e.g. cloudwatch) 
   workspaces:
   - name: source
+    mountPath: /src/k8s.io/
   - name: results
   steps:
   - name: git-clone      
@@ -67,11 +70,36 @@ spec:
       # we are not testing PVS at this point
       CL2_ENABLE_PVS: false
       PROMETHEUS_SCRAPE_KUBE_PROXY: true
+      PROMETHEUS_SCRAPE_APISERVER_ONLY: true
       ENABLE_SYSTEM_POD_METRICS: false
       NODE_MODE: master 
       EOL
       cat $(workspaces.source.path)/overrides.yaml
       cp $(workspaces.source.path)/overrides.yaml $(workspaces.results.path)/overrides.yaml
+
+      # TODO: Currently pathing the prometheus manifests for remote write. Move this to upsteam going forward.
+      echo "volumeBindingMode: WaitForFirstConsumer" >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+      cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+      cat << EOF >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+        containers:
+          - name: aws-sigv4-proxy-sidecar
+            image: public.ecr.aws/aws-observability/aws-sigv4-proxy:1.0
+            args:
+              - --name
+              - aps
+              - --region
+              - $(params.region)
+              - --host
+              - aps-workspaces.$(params.region).amazonaws.com
+              - --port
+              - :8005
+            ports:
+              - name: aws-sigv4-proxy
+                containerPort: 8005
+        remoteWrite:
+          - url: http://localhost:8005/workspaces/$(params.amp-workspace-id)/api/v1/remote_write
+      EOF
+      cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
   - name: run-loadtest
     image: public.ecr.aws/kit/clusterloader2:0213bea
     onError: continue
--- a/tests/tasks/generators/clusterloader/load.yaml
+++ b/tests/tasks/generators/clusterloader/load.yaml
@@ -36,7 +36,8 @@ spec:
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
   - name: amp-workspace-id
-    description: The remote amp workspace id where remote needs to happen.
+    description: The AMP workspace ID where remote write needs to happen.
+    default: ""
   results:
     - name: datapoint
       description: Stores the CL2 result that can be consumed by other tasks (e.g. cloudwatch) 
@@ -76,7 +77,9 @@ spec:
       EOL
       cat $(workspaces.source.path)/overrides.yaml
       cp $(workspaces.source.path)/overrides.yaml $(workspaces.results.path)/overrides.yaml
-
+      
+      # Enable Prometheus if the remote workspace id is provided
+      if [ -n "$(params.amp-workspace-id)" ]; then
       # TODO: Currently pathing the prometheus manifests for remote write. Move this to upsteam going forward.
       echo "volumeBindingMode: WaitForFirstConsumer" >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
       cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
@@ -100,6 +103,7 @@ spec:
           - url: http://localhost:8005/workspaces/$(params.amp-workspace-id)/api/v1/remote_write
       EOF
       cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+      fi
   - name: run-loadtest
     image: public.ecr.aws/kit/clusterloader2:0213bea
     onError: continue
@@ -109,9 +113,12 @@ spec:
       if [ -n "$(params.endpoint)" ]; then
         ENDPOINT_FLAG="--endpoint $(params.endpoint)"
       fi
+      if [ -n "$(params.amp-workspace-id)" ]; then
+        CL2_PROMETHEUS_FLAGS="--enable-prometheus-server=true --prometheus-storage-class-provisioner kubernetes.io/aws-ebs --prometheus-storage-class-volume-type gp2 --prometheus-manifest-path=$(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/"
+      fi
       aws eks $ENDPOINT_FLAG update-kubeconfig --name $(params.cluster-name) --region $(params.region)
       cat $(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml
-      ENABLE_EXEC_SERVICE=false /clusterloader --kubeconfig=/root/.kube/config --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr --v=2
+      ENABLE_EXEC_SERVICE=false /clusterloader --kubeconfig=/root/.kube/config --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr --v=2 $CL2_PROMETHEUS_FLAGS
       if [ $? -eq 0 ]; then
       echo "1" | tee $(results.datapoint.path)
       else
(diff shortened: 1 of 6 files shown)


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -67,11 +70,36 @@ spec:
...
       EOL
       cat $(workspaces.source.path)/overrides.yaml
       cp $(workspaces.source.path)/overrides.yaml $(workspaces.results.path)/overrides.yaml
+
+      # TODO: Currently pathing the prometheus manifests for remote write. Move this to upsteam going forward.
+      echo "volumeBindingMode: WaitForFirstConsumer" >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+      cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+      cat << EOF >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
+        containers:
+          - name: aws-sigv4-proxy-sidecar
```
mengqiy, line 85: Sidecar works. But have you considered using the built-in support of SigV4 which is available since [2.26.0](https://aws.amazon.com/blogs/opensource/prometheus-2-26-0-adds-aws-signature-version-4-support/).
Ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/apis/monitoring/v1/types.go#L1074-L1088
IMO using the prometheus operator's API may make it easier to upstream the change.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -77,15 +77,16 @@ spec:
...
       done
       remaining_nodes=$((nodes%1000))
       if [[ $remaining_nodes -gt 0 ]]
       then
         echo "The remaining_nodes var is greater than 0."
-        create_dp 0 $remaining_nodes
+        create_dp 0 $remaining_nodes c5.large,m5.large,r5.large,t3.large,t3a.large,c5a.large,m5a.large,r5a.large
       fi
-      
+      # Creating an extra asg with 1 node for prometheus server pod scheduling.
```
mengqiy, line 88: How do you ensure only Prometheus server (and operator) can run on this node?

tests/tasks/generators/clusterloader/load.yaml
mengqiy: This comment doesn't block this PR from merging as long as we still use AMP and AMG.
I'm OK to address this in a followup PR if needed.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -77,15 +77,16 @@ spec:
       for i in $(seq 1 $asgs)
       do
         #max number of nodes MNG allows per ASG
-        create_dp $i 1000
+        create_dp $i 1000 c5.large,m5.large,r5.large,t3.large,t3a.large,c5a.large,m5a.large,r5a.large
```
hakuna-matatah, line 80: Could you parametrize these and default it to these instance types ? 
Given you were running into icing issues when you tried it with different set of instances when testing prom changes  -  It will be much easier to control the mixed-instance-group dynamically.

hakuna-matatah:
could you add permissions/policies required to perform AMP operations here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L129-L148 so it works out of box without manual intervention during setup.

Also, could you add a readme to setup AMP as it's a pre-req for some of our tekton pipelines in our doc [here](https://github.com/awslabs/kubernetes-iteration-toolkit/tree/main/tests)

hakuna-matatah:
Very nice to see all this coming through. Thank you Ashish.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -35,11 +35,14 @@ spec:
...
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  - name: amp-workspace-id
+    description: The remote amp workspace id where remote needs to happen.
   results:
     - name: datapoint
       description: Stores the CL2 result that can be consumed by other tasks (e.g. cloudwatch) 
   workspaces:
   - name: source
+    mountPath: /src/k8s.io/
```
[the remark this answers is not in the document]
ashishranjan738: I just preferred this path as it coincides with GOPATH and also some hardcoding are there in CL2 code regarding this path https://github.com/kubernetes/perf-tests/blob/release-1.23/clusterloader2/pkg/prometheus/prometheus.go#L66. However if we are passing the manifests explicitly which indeed we are, we can avoid this. But this shouldn't be a breaking change.

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: yeah thats the plan. i tried upgrading the prometheus and prometheus operator, but with the new versions available in the gcr repository the prometheus pods fails to come up with error in parsing configmap. I will try to raise a fix to this upstream and we can remove this patch going forward, but we can keep this atleast till 1.26.

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: So by default the cl2 test allots cpu and memory to prometheus based on the number of nodes the cluster has https://github.com/kubernetes/perf-tests/blob/release-1.23/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L19. Instead of having a specific node just to run the prometheus pod, what I have tried to ensure is we have at least 1 node which satisfies the cpu and memory request of the prometheus. Definitely, we can have a more advance logic to reserve one node based on taints going forward, but I feel this more simple currently.  Let me know your thoughts on this ?

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: true..have added this.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -35,11 +35,14 @@ spec:
     description: "aws eks enpoint to create clusters against"
   - name: cluster-name
     description: The name of the EKS cluster you want to spin.
+  - name: amp-workspace-id
+    description: The remote amp workspace id where remote needs to happen.
```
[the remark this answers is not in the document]
ashishranjan738: done.

tests/tasks/generators/clusterloader/load.yaml
mengqiy: What prometheus version does cluster-loader currently use?

ashishranjan738: v2.25 https://github.com/kubernetes/perf-tests/blob/release-1.23/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L50

tests/tasks/setup/eks/awscli-mng.yaml
mengqiy: Prometheus server pod may require a very large node when nodes in the cluster are a lot. Have you seen it failing to be scheduled due to nodes are too small?
Some pods under test may end up be scheduled on this large node and it may some inaccuracy in our test.
This is not a show stopper. But we don't plan to address it now. We should track it (at least a TODO).
@hakuna-matatah Thoughts?

ashishranjan738: @mengqiy yeah initially when i tried it was failing due to small nodes.. i had to explicitly adjust my nodes from node groups to add a big schedulable node to my cluster for prometheus.  Current finding is for 10k cluster 1 m5.4xlarge is enough to schedule a pod for Prometheus..maybe we can have a logic to dynamically add one instance based on the cluster nodes and prometheus cpu and memory pre estimation. But yeah open for thoughts on this. +1 to tracking. I will create an issue for this.

ashishranjan738:
> could you add permissions/policies required to perform AMP operations here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L129-L148 so it works out of box without manual intervention during setup.
@hakuna-matatah how is adding to place gonna help us ? We need this permission to the role we pass in the mng task https://github.com/awslabs/kubernetes-iteration-toolkit/blob/cf70d5c5483bc1480a79693d3757b8f379b6f3b5/tests/tasks/setup/eks/awscli-mng.yaml#L28 . i guess this is currently not generated via cdk

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: can you add a backlog issue for tracking this @ashishranjan738 and assign a label as 1.26.

tests/tasks/setup/eks/awscli-mng.yaml
hakuna-matatah: You mean, you are going to add it ?

tests/tasks/setup/eks/awscli-mng.yaml
hakuna-matatah: Given cl2 prom setup is more robust in requesting the [resources](https://github.com/kubernetes/perf-tests/blob/release-1.23/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L19) we can punt adding taints to future if you think it's requires more work. Please cut a backlog as you see fit.
But I generally agree with David, we should be not add any load-test pods to node because there is a risk of prometheus pods getting evicted or OOM killed/restarted which not only it will cost us loosing metrics data but it might also fail our tests because of it's potential impact on APIServer SLOs measurements.

hakuna-matatah:
> > > could you add permissions/policies required to perform AMP operations here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L129-L148 so it works out of box without manual intervention during setup.
> > 
> > 
> > @hakuna-matatah how is adding to place gonna help us ? We need this permission to the role we pass in the mng task which i guess this is currently not generated via cdk
> > https://github.com/awslabs/kubernetes-iteration-toolkit/blob/cf70d5c5483bc1480a79693d3757b8f379b6f3b5/tests/tasks/setup/eks/awscli-mng.yaml#L28
> 
> No, tekton pods leverage IRSA hence the code link i shared in my comment will work for us if you add it there. The one you are pointing to is a node role which MNG APIs need as param for creating MNGs.

hakuna-matatah:
> > > could you add permissions/policies required to perform AMP operations here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L129-L148 so it works out of box without manual intervention during setup.
> > 
> > 
> > @hakuna-matatah how is adding to place gonna help us ? We need this permission to the role we pass in the mng task which i guess this is currently not generated via cdk
> > https://github.com/awslabs/kubernetes-iteration-toolkit/blob/cf70d5c5483bc1480a79693d3757b8f379b6f3b5/tests/tasks/setup/eks/awscli-mng.yaml#L28
> 
~> No, tekton pods leverage IRSA hence the code link i shared in my comment will work for us if you add it there. The one you are pointing to is a node role which MNG APIs need as param for creating MNGs.~

Oops accidentally closed instead of adding commenting.
Synced with Ashish on this. Here in this case,  prom pod on guest cluster needs access, we are going to go with node role creation tekton step/task for this and wrap all policies required inside it for guest clusters.

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: yeah created here https://github.com/awslabs/kubernetes-iteration-toolkit/issues/326

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: Makes sense. I will create a follow up PR for this to have a specific node for prometheus. Thanks @mengqiy @hakuna-matatah

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: yeah pushed in a different commit.

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: cool, thanks. I added label 1.26 to it.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -49,21 +49,27 @@ spec:
...
-      nodes=$(params.desired-nodes)
+      nodes=$(($(params.desired-nodes)-1))
       asgs=$((nodes/1000))
       asg_name=$(params.cluster-name)-nodes
       create_dp()
       {
         CREATED_NODEGROUP=$(aws eks $ENDPOINT_FLAG --region $(params.region) list-nodegroups --cluster-name $(params.cluster-name)  --query 'nodegroups[?@==`'$asg_name-$1'`]' --output text)
-        #ToDo: paramterize instance-types
+        echo $3
+        EC2_INSTANCES=$3
```
hakuna-matatah, line 59: oh! by this comment [here](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/325#discussion_r1036659368) I meant, adding as tekton param so we can dynamically alter it for pipeline/task runs as and when we get iced.

ashishranjan738: okay..I will add this with prometheus taints PR.

hakuna-matatah (dismissed):
LGTM!
/Do-not-merge

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: Have added this logic to a different comment.

mengqiy (approved):

mengqiy:
Please work with @hakuna-matatah to understand why he added `/Do-not-merge` in the comment above.

PR #327, opened 2022-12-05
Taints & Node Selector for Prometheus, Cluster Role Tasks and follow ups for #325

ashishranjan738:
This PR introduces node selector and node taint patch for scheduling CL2 prometheus and also adds other minor fixes discussed in PR #325 like parameterizing instance types in load test, cluster role creation tasks, moving cw-metrics file to addons and adding cluster-name label in prometheus.


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -83,6 +86,8 @@ spec:
       # TODO: Currently pathing the prometheus manifests for remote write. Move this to upsteam going forward.
       echo "volumeBindingMode: WaitForFirstConsumer" >> $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
       cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/0ssd-storage-class.yaml
+      sed -i '/nodeSelector/d' $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
```
mengqiy, line 89: Please add comment explaining what you are trying to do here.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -83,16 +86,16 @@ spec:
...
       remaining_nodes=$(((nodes)%1000))
       if [[ $remaining_nodes -gt 0 ]]
       then
         echo "The remaining_nodes var is greater than 0."
-        create_dp 0 $remaining_nodes
+        create_dp 0 $remaining_nodes "$(params.mng-host-instance-types)"
       fi
       # Creating an extra asg with 1 large node to ensure prometheus server has a scheduling space.
-      create_dp $(($asgs + 1)) 1 m5.4xlarge
+      create_dp "monitoring" 1 "$(params.monitoring-host-instance-types)" "key=monitoring,value=true,effect=NO_SCHEDULE"
```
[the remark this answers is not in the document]
ashishranjan738: I can address this in role creation task PR.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -119,17 +130,21 @@ spec:
...
       else
       echo "0" | tee $(results.datapoint.path)
       fi
+      exit $exit_code
     timeout: 30000s
   - name: upload-results
     image: amazon/aws-cli
     workingDir: $(workspaces.results.path)
     script: |
+      S3_RESULT_PATH=$(cat $(workspaces.source.path)/s3_result_path.txt)
```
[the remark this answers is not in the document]
ashishranjan738: So the idea here is to add s3 result path as prometheus label for easy visualization of metrics. Since the result path is computed dynamically in the Tekton task using timestamp https://code.amazon.com/packages/EKSScalabilityTektonTestRuns/blobs/685d1c87e58fd4525ae50c177ee12301ff58ebfa/--/loadtests/triggers/eks/awscli-5k-load.yaml#L88. We are computing the s3 result path here and the same is passed to the next step so that the same path is used there also. Recomputing the result path on next step can result in a different path so it passed via file.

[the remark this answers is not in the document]
mengqiy: I know we use timestamp in `params.results-bucket`. But I expect it to be fixed once initialized.
Are you saying `params.results-bucket` is not fixed. I'd surprised if that's the case.

tests/tasks/setup/eks/awscli-mng.yaml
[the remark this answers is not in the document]
ashishranjan738: updated the logic.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -27,6 +27,12 @@ spec:
     default: ""
   - name: host-cluster-node-role-arn
     description: arn of the hostcluster node role. This tightly coupled to code here  - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3ed1bbd47f7b8f111208e977acaa3edfa1834ca8/substrate/pkg/controller/substrate/cluster/addons/karpenter.go#L52 so if it's changed there, it should be changed here. This helps us to avoid creating a separate noderole for nodegroups.
+  - name: mng-host-instance-types
+    description: Instance types for managed nodes.
+    default: "c5.large m5.large r5.large t3.large t3a.large c5a.large m5a.large r5a.large"
+  - name: monitoring-host-instance-types
```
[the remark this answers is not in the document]
ashishranjan738: updated the logic.

tests/tasks/generators/clusterloader/load.yaml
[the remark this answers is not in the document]
ashishranjan738: have started tekton result object for passing the values to next step instead of file.

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -59,6 +59,9 @@ spec:
     image: alpine/k8s:1.22.6
     workingDir: $(workspaces.source.path)
     script: |
+      S3_RESULT_PATH=$(params.results-bucket)
+      echo $S3_RESULT_PATH > $(workspaces.source.path)/s3_result_path.txt
```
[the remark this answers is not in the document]
ashishranjan738: have started tekton result object for passing the values to next step instead of file.

tests/tasks/generators/clusterloader/load.yaml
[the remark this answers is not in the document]
mengqiy: So you are saying `$(date +%s)` will be evaluated in every step that uses it. This is interesting to learn.

>have started tekton result object for passing the values to next step instead of file.

Yeah, that's something I'd suggest. It's definitely light-weight than communicating through s3.
We are using it in the CloudWatch emitting metrics tasks as well.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -19,14 +19,23 @@ spec:
...
+    description: The minimum number of nodes in the cluster nodegroup.
   - name: max-nodes
     default: "1000"
-    description: The maximum number of nodes in the cluster.
+    description: The maximum number of nodes in the cluster nodegroup.
   - name: endpoint
     default: ""
   - name: host-cluster-node-role-arn
     description: arn of the hostcluster node role. This tightly coupled to code here  - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3ed1bbd47f7b8f111208e977acaa3edfa1834ca8/substrate/pkg/controller/substrate/cluster/addons/karpenter.go#L52 so if it's changed there, it should be changed here. This helps us to avoid creating a separate noderole for nodegroups.
+  - name: mng-host-instance-types
```
hakuna-matatah, line 30: nit: to keep it consistent, can you strip `mng` from the name ? Given it's already a task related to mng and also you don't use `mng` in name for `host-taints` param.

tests/tasks/generators/clusterloader/load.yaml
mengqiy: Please add these comment in the file for posterity.
We want other people not just 3 of us to understand why we are doing it when they look at this file.
We should be able to understand it after we come back months later.

tests/tasks/setup/eks/awscli-role.yaml
```diff
@@ -0,0 +1,61 @@
+---
```
hakuna-matatah, line 1: I would see this task to more generic enough like as simple as doing any role creation, doesn't have to be specific to eks service role creation (or) eks node role creation. Makes sense ? Could we abstract that out ? Basically a task which takes input to create a specific role using a path to the policies.

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: yeah I can replace it.

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: yeah added.

tests/tasks/setup/eks/awscli-role.yaml
ashishranjan738: Right..as discussed offline I have made this change in the new commit.

tests/tasks/teardown/awscli-eks.yaml
```diff
@@ -37,6 +39,13 @@ spec:
           aws eks wait nodegroup-deleted --nodegroup-name $i --cluster-name $(params.cluster-name) $ENDPOINT_FLAG --region $(params.region);
       done;
       aws eks delete-cluster --name $(params.cluster-name) --region $(params.region) $ENDPOINT_FLAG
+  - name: teardown-eks-role-stack
+    image: alpine/k8s:1.23.13
+    script: |
+      echo "Approving KCM requests"
+      kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
```
mengqiy, line 46: I don't see a need to approve CSRs.
There are lightly to have tens of thousands of CSR accumulated over the time that the test are run. Each kubectl will create multiple. It can take a long time to approve all of them.

tests/assets/eks_node_role_policy.json
```diff
@@ -0,0 +1,167 @@
+{
```
mengqiy, line 1: I don't see this file being used anywhere. Did I miss anything?

tests/assets/eks_service_role_trust_policy.json
```diff
@@ -0,0 +1,16 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Principal": {
+                "Service": [
+                    "eks-gamma.aws.internal",
+                    "eks.amazonaws.com",
+                    "eks-beta.aws.internal"
```
mengqiy, line 10: @hakuna-matatah Any concerns? I'd defer to you.

tests/assets/eks_service_role_policy.json
```diff
@@ -0,0 +1,5 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+    ]
```
mengqiy, line 4: Is this just a placeholder?

tests/assets/eks_node_role_trust_policy.json
```diff
@@ -0,0 +1,15 @@
+{
```
mengqiy, line 1: Same question here.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
```diff
@@ -28,34 +34,119 @@ spec:
     taskRef:
       kind: Task
       name:  slack-notification
+  - name: create-cluster-service-role
+    params:
+    - name: stack-name
+      value: $(params.cluster-name)-service-role
+    - name: role-cfn-url
+      value: $(params.service-role-cfn-url)s
```
mengqiy, line 42: `s` at the end? typo?

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
```diff
@@ -28,34 +34,119 @@ spec:
...
+            description: holds the preferred node role Arn
+        params:
+          - name: cfn-service-role-arn
+          - name: cfn-node-role-arn
+          - name: param-host-cluster-node-role-arn
+          - name: param-service-role-arn
+        steps:
+          - name: get-prefered-role
+            image: alpine
+            script: |
```
mengqiy, line 90: Any reason why we prefer an embeded one over ref here?

tests/tasks/teardown/awscli-eks.yaml
ashishranjan738: yeah makes sense.. Have removed it.

tests/assets/eks_service_role_policy.json
ashishranjan738: Actually it was getting used earlier and I forgot to remove it. Have deleted the un needed jsons.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
ashishranjan738: Unintentional change..I guess it got in by mistake.

ashishranjan738: removed it.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
ashishranjan738: I am open to thoughts here..the reason why i kept it embedded is it basically contains some pipeline orchestration logic to decide on which ARN to use. Since this orchestration logic is strictly tied with pipelines and doesn't have any meaning independently i wanted to keep it coupled with the pipeline.

tests/assets/eks_node_role_trust_policy.json
ashishranjan738: removed it..not getting used.

tests/assets/eks_service_role_trust_policy.json
ashishranjan738: this file is not getting used so removed it.

tests/assets/eks_node_role_policy.json
ashishranjan738: removed it as it is not getting used.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
mengqiy: My only concern here is we are duplicating it here and `tests/pipelines/eks/awscli-eks-cl2-load.yaml`. It may get out-of-sync.
@hakuna-matatah WDYT?

tests/tasks/setup/eks/awscli-role.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - name: stack-name
+    description: The name of the Role name you want to spin.
+  - name: role-cfn-url
+    description: The url of the CFN YAML/JSON to create IAM Roles 
+  - name: region
+    default: "us-west-2"
+  - name: role-name
+    description: The name of the role that needs to be created.
+  steps:
+  - name: create-vpc
```
hakuna-matatah, line 23: wrong step name. May be copy paste error ? :D

tests/pipelines/eks/awscli-eks-cl2-load.yaml
```diff
@@ -11,8 +11,10 @@ spec:
   - name: cluster-name
   - name: endpoint
   - name: servicerole
+    default: ""
```
hakuna-matatah, line 14: you can remove this at pipeline level, given it's already a default at task level IIUC.

tests/pipelines/eks/awscli-eks-cl2-load.yaml
```diff
@@ -11,8 +11,10 @@ spec:
...
   - name: desired-nodes
   - name: host-cluster-node-role-arn
+    default: ""
```
hakuna-matatah, line 17: same with this, you can remove this at pipeline level as we are anyway using this is task level default.

tests/pipelines/eks/awscli-eks-cl2-load.yaml
```diff
@@ -43,35 +49,117 @@ spec:
...
+    params:
+    - name: stack-name
+      value: $(params.cluster-name)-node-role
+    - name: role-cfn-url
+      value: $(params.node-role-cfn-url)
+    - name: role-name
+      value: "$(params.cluster-name)-node-role"
+    taskRef:
+      kind: Task
+      name: awscli-eks-role-create
```
hakuna-matatah, line 73: drop `eks` from task name. Just `awscli-role-create` as it's a generic task.

hakuna-matatah, line 73: drop `eks` from task name. As it's a generic task.

tests/pipelines/eks/awscli-eks-cl2-load.yaml
```diff
@@ -43,35 +49,117 @@ spec:
...
+    params:
+    - name: stack-name
+      value: $(params.cluster-name)-service-role
+    - name: role-cfn-url
+      value: $(params.service-role-cfn-url)
+    - name: role-name
+      value: "$(params.cluster-name)-service-role"
+    taskRef:
+      kind: Task
+      name: awscli-eks-role-create
```
hakuna-matatah, line 62: same as other comment

tests/pipelines/eks/awscli-eks-cl2-load.yaml
ashishranjan738: removing this from here will not allow us to trigger runs without empty params

tests/pipelines/eks/awscli-eks-cl2-load.yaml
ashishranjan738: same here

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
hakuna-matatah: Yeah, I'm inclined towards not having to duplicate this logic across pipelines. I would rather suggest to make it less coupling, read `service-role-name` as param instead of relying on `arn` and try to fetch the `role-arn` with in the task based on the name supplied [
here](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/327/files#diff-566fabd5277b56d45d422840baf5a353ec32543ea2f755ad6aa45eca934830deR40-R43) 

Same for node-role-arn too.

tests/tasks/setup/eks/awscli-role.yaml
ashishranjan738: yeah..updated

tests/pipelines/eks/awscli-eks-cl2-load.yaml
ashishranjan738: done

tests/pipelines/eks/awscli-eks-cl2-load.yaml
ashishranjan738: done.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
ashishranjan738: okay..I have updated the tasks and pipelines to accept role names and currently we have assumptions in the task for default role name based on the cluster role name.

tests/pipelines/eks/awscli-eks-cl2-load.yaml
ashishranjan738: as discussed offline removed this. The user needs to explicitly pass role names for these pipelines.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
```diff
@@ -18,6 +16,10 @@ spec:
   - name: slack-message
   - name: amp-workspace-id
     default: ""
```
hakuna-matatah, line 18: why do we need this `default` ?

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
```diff
@@ -28,34 +30,83 @@ spec:
...
       value: $(params.cluster-name)
-    - name: servicerole
-      value: $(params.servicerole)
+    - name: service-role-name
+      value: $(params.cluster-name)-service-role
     - name: endpoint
       value: $(params.endpoint)
     runAfter:
-    - slack-notification
+    - create-cluster-node-role
```
hakuna-matatah, line 68: this should be `create-cluster-service-role` as we need service role for `create-cluster` task

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -101,6 +109,16 @@ spec:
                 containerPort: 8005
         remoteWrite:
           - url: http://localhost:8005/workspaces/$(params.amp-workspace-id)/api/v1/remote_write
+            queueConfig:
+              capacity: 2500
```
hakuna-matatah, line 113: why do we need this ? Defaults are not enough for us ?

hakuna-matatah: can we document why we need these values ?

tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -119,17 +137,21 @@ spec:
...
       cat $(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml
       ENABLE_EXEC_SERVICE=false /clusterloader --kubeconfig=/root/.kube/config --testconfig=$(workspaces.source.path)/perf-tests/clusterloader2/testing/load/config.yaml --testoverrides=$(workspaces.source.path)/overrides.yaml --nodes=$(params.nodes) --provider=eks --report-dir=$(workspaces.results.path) --alsologtostderr --v=2 $CL2_PROMETHEUS_FLAGS
-      if [ $? -eq 0 ]; then
+      exit_code=$?
+      if [ $exit_code -eq 0 ]; then
       echo "1" | tee $(results.datapoint.path)
       else
       echo "0" | tee $(results.datapoint.path)
       fi
+      exit $exit_code
```
hakuna-matatah, line 146: Thanks for fixing this.

tests/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -19,14 +19,23 @@ spec:
...
   - name: max-nodes
     default: "1000"
-    description: The maximum number of nodes in the cluster.
+    description: The maximum number of nodes in the cluster nodegroup.
   - name: endpoint
     default: ""
-  - name: host-cluster-node-role-arn
-    description: arn of the hostcluster node role. This tightly coupled to code here  - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3ed1bbd47f7b8f111208e977acaa3edfa1834ca8/substrate/pkg/controller/substrate/cluster/addons/karpenter.go#L52 so if it's changed there, it should be changed here. This helps us to avoid creating a separate noderole for nodegroups.
+  - name: host-cluster-node-role-name
+    description: name of the hostcluster node role. This tightly coupled to code here  - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3ed1bbd47f7b8f111208e977acaa3edfa1834ca8/substrate/pkg/controller/substrate/cluster/addons/karpenter.go#L52 so if it's changed there, it should be changed here. This helps us to avoid creating a separate noderole for nodegroups.
```
hakuna-matatah, line 29: This description may not be valid anymore especially for `EKS Managed cluster` ?

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
ashishranjan738: we don't need it..we can get rid of it..I kept it initially to make so that it doesn't break the existing runs of our pipelines..but since now we are going to anyway update it we can get rid of it.

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
ashishranjan738: we can have both..we can do the stack setup in parallel to save some time. :)

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: i read an aws prometheus blog and they suggested these values. So better to keep this.

tests/tasks/setup/eks/awscli-mng.yaml
ashishranjan738: updated.

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: https://aws.amazon.com/blogs/mt/getting-started-amazon-managed-service-for-prometheus/

tests/pipelines/eks/awscli-cl2-load-with-addons.yaml
hakuna-matatah: Yes, we can have both too. As long as we have service role created before we can save race condition failure mode. Discussed offline on this.

hakuna-matatah (approved):
LGTM!

PR #315, opened 2022-11-04
[chore] Update EKS-D latest image tags

premdass:
Updating latest image tags of EKS-D

Towards https://github.com/aws/eks-distro/issues/1174

Issue #, if available:
https://github.com/aws/eks-distro/issues/1174

Description of changes:
Updating the EKS-D image tags to latest

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #328, opened 2022-12-06
[DO-NOT-MERGE] fix EKS-D image tags

mengqiy:
`kubeVersionXXXTag`s point to invalid image tags. They are fixed in this PR.
Also update the image tags for etcd, coredns and aws-iam-authenticator to the latest according to EKS-D.

Fixed the issues that were introduced in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/315.


hakuna-matatah (approved):
LGTM

PR #329, opened 2022-12-07
grant AmazonPrometheusRemoteWriteAccess in kit operator

mengqiy:
The kit guest cluster can't do remote write w/o this permission.
Attach AmazonPrometheusRemoteWriteAccess policy. The prometheus server can't do remote write w/o this change.


ashishranjan738:
LGTM since KIT guest cluster uses role with policy https://github.com/awslabs/kubernetes-iteration-toolkit/blob/a6763db526664d3d509cb75914d4059b91b8daff/operator/pkg/awsprovider/iam/reconciler.go#L43 It needs to be updated.

hakuna-matatah (approved):
LGTM

PR #334, opened 2022-12-13
bug fixes for test failures

hakuna-matatah:
Issue #, if available:
I had these local bug fixes for a while now to fix some of the failures we have been seeing in our tests in the past.

Description of changes:

### DIFF
--- a/tests/tasks/generators/clusterloader/load.yaml
+++ b/tests/tasks/generators/clusterloader/load.yaml
@@ -62,7 +62,7 @@ spec:
       cat > "$(workspaces.source.path)/overrides.yaml" <<EOL
       NODES_PER_NAMESPACE: $(params.nodes-per-namespace)
       CL2_LOAD_TEST_THROUGHPUT: $(params.cl2-load-test-throughput)
-      CL2_SCHEDULER_THROUGHPUT_THRESHOLD: 20
+      CL2_SCHEDULER_THROUGHPUT_THRESHOLD: 70
       PODS_PER_NODE: $(params.pods-per-node)
       CL2_USE_HOST_NETWORK_PODS: false
       # we are not testing statefulsets at this point
@@ -105,7 +105,7 @@ spec:
       cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
       fi
   - name: run-loadtest
-    image: public.ecr.aws/kit/clusterloader2:0213bea
+    image: 197575167141.dkr.ecr.us-west-2.amazonaws.com/clusterloader2:latest
     onError: continue
     script: |
       #!/bin/bash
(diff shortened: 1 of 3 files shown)


tests/tasks/generators/clusterloader/load.yaml
```diff
@@ -105,7 +105,7 @@ spec:
       cat $(workspaces.source.path)/perf-tests/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml
       fi
   - name: run-loadtest
-    image: public.ecr.aws/kit/clusterloader2:0213bea
+    image: 197575167141.dkr.ecr.us-west-2.amazonaws.com/clusterloader2:latest
```
ashishranjan738, line 108: why is this changed ? is this a version upgrade ?

tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
```diff
@@ -51,6 +51,10 @@ spec:
       aws eks $ENDPOINT_FLAG --region $(params.region) wait cluster-active --name $(params.cluster-name) 
       aws eks $ENDPOINT_FLAG update-kubeconfig --name $(params.cluster-name) --region $(params.region)
       # enable PD on the cluster 
+      kubectl api-versions 
+      kubectl api-resources
+      # To do - remove this, temporarily adding because apparently apiserver is not recongnizing deamonset for sometime inconsistently, need to see if livez and readyz are active
```
ashishranjan738, line 56: shouldn't we print the resources after sleep then ? so that get we can get better info on whether the resources are ready or not ?

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: I have applied an internal patch to fix the failures we are seeing for EKS scale tests. I have a PR in upstream and in the process to make it to upstream.

hakuna-matatah: we have been using this patch for a while now.

tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
hakuna-matatah: After 5mins, it unlikely any of the resources are not ready, I'm mostly interested before sleep because ideally CFN-wait should wait until cluster is ready. Does it makes sense ?

tests/tasks/generators/clusterloader/load.yaml
ashishranjan738: okay. you can link that PR here for tracking.

tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
ashishranjan738: got it

tests/tasks/generators/clusterloader/load.yaml
hakuna-matatah: https://github.com/kubernetes/perf-tests/pull/2178

natherz97 (approved):

PR #336, opened 2022-12-13
Fix incorrect name for nodegroup create task

ashishranjan738:
This commit fixes incorrect name for nodegroup task.


hakuna-matatah (approved):
LGTM

PR #341, opened 2022-12-16
Adds flag to pass instance type during bootstrap

ashishranjan738:
This commit adds flag to pass on instance type during bootstrap for the kit management cluster.

### DIFF
--- a/substrate/cmd/kitctl/root.go
+++ b/substrate/cmd/kitctl/root.go
@@ -54,14 +54,13 @@ func newRootCmd(args []string) *cobra.Command {
 		Long: `kitctl help users provision Kubernetes clustes using kit-operator.
 		It also configures the cloud provider environment to get started easily`,
 	}
-	flags := cmd.PersistentFlags()
-	options.addFlags(flags)
-	if err := flags.Parse(args); err != nil {
-		panic(err)
-	}
+
 	// Add subcommands
 	cmd.AddCommand(bootstrapCommand())
 	cmd.AddCommand(deleteCommand())
+
+	flags := cmd.PersistentFlags()
+	options.addFlags(flags)
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


substrate/cmd/kitctl/root.go
```diff
@@ -68,13 +68,15 @@ func newRootCmd(args []string) *cobra.Command {
 var options = &Options{}
 
 type Options struct {
-	file  string
-	debug bool
-	help  bool
+	file         string
+	debug        bool
+	help         bool
+	instanceType string
```
prateekgogia, line 74: I think this will also apply to delete command which doesn't make sense for the user and can get confusing. Right place for this IMO, for this flag is for subcommand `bootstrap`. please correct me if I am wrong.

prateekgogia (changes requested):

substrate/cmd/kitctl/root.go
ashishranjan738: hmm makes sense..we can have a local flag on the bootstrap cmd

ashishranjan738: have update it to local flag.

substrate/cmd/kitctl/root.go
```diff
@@ -54,14 +54,13 @@ func newRootCmd(args []string) *cobra.Command {
 		Long: `kitctl help users provision Kubernetes clustes using kit-operator.
 		It also configures the cloud provider environment to get started easily`,
 	}
-	flags := cmd.PersistentFlags()
-	options.addFlags(flags)
-	if err := flags.Parse(args); err != nil {
-		panic(err)
-	}
```
prateekgogia, line 61: Do we not need `flags.Parse` anymore?

ashishranjan738: no.

PR #350, opened 2023-01-13
Tekton task to clean up the vpc resources

Ganiredi:
Issue #, if available:

Description of changes:

This task will delete the CFN stack if stack  exists with the name provided.The following two images have a CFN stack that was created and deleted .

## CFN Stack Created
Screen Shot 2023-01-13 at 11 49 37 AM

## CFN Stack Deleted
Screen Shot 2023-01-13 at 11 52 49 AM

## Tekton Task Output 
image

### DIFF
--- a/tests/tasks/teardown/awscli-vpc-delete.yaml
+++ b/tests/tasks/teardown/awscli-vpc-delete.yaml
@@ -1,0 +1,34 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: awscli-delete-vpc
+  namespace: tekton-pipelines
+spec:
+  description: |
+    This Task can be used to delete CloudFormation stack containing VPC resources that was used for EKS clusters.
+  params:
+    - name: stack-name
+      description: The name of the VPC name you want to delete.
+    - name: region
+      default: "us-west-2"
+  steps:
+    - name: awscli-delete-vpc
+      image: alpine/k8s:1.22.6
+      script: |
+        #!/bin/bash
+        echo "Approving KCM requests"
+        kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
+        aws sts get-caller-identity 
+        # Check if the stack exists
+        aws cloudformation --region $(params.region) describe-stacks --stack-name $(params.stack-name)
+        if [ $? -ne 0 ]; then
+          echo "Stack $(params.stack-name) not found. Exiting..."
+          exit 1
+        else
+          echo "Deleting stack $(params.stack-name)..."
+        fi
+        #Deletes the CFN stack
+        aws cloudformation delete-stack --region $(params.region) --stack-name $(params.stack-name) 
+        # Wait for the stack to be deleted
+        aws cloudformation wait stack-delete-complete --region $(params.region) --stack-name $(params.stack-name)
+        echo "Stack deleted successfully!"
--- a/tests/assets/amazon-eks-vpc.json
+++ b/tests/assets/amazon-eks-vpc.json
@@ -664,6 +664,7 @@
         },
         "PublicSubnet01": {
             "Type": "AWS::EC2::Subnet",
+            "DependsOn": "VPC",
             "Metadata": {
                 "Comment": "Subnet 01"
             },
@@ -701,6 +702,7 @@
         },
         "PublicSubnet02": {
             "Type": "AWS::EC2::Subnet",
+            "DependsOn": "VPCCIDRBlock2",
             "Metadata": {
                 "Comment": "Subnet 02"
             },
@@ -738,6 +740,7 @@
         },
         "PublicSubnet03": {
             "Condition": "HasMoreThan2Azs",
+            "DependsOn": "VPCCIDRBlock3",
             "Type": "AWS::EC2::Subnet",
             "Metadata": {
                 "Comment": "Subnet 03"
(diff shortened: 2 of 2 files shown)


tests/tasks/teardown/awscli-vpc-delete.yaml
```diff
@@ -0,0 +1,32 @@
...
+          #!/bin/bash
+          echo "Approving KCM requests"
+          kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
+          aws sts get-caller-identity 
+          # Check if the stack exists
+          stack_exists=$(aws cloudformation --region $(params.region) describe-stacks --stack-name $(params.stack-name) 2>&1)
+          if [[ $stack_exists == *"does not exist"* ]]; then
+            echo "Error: Stack $(params.stack-name) does not exist."
+            exit 1
+          fi
```
hakuna-matatah, line 27: I hesitate to depend on the `error message` instead of `error code`. 
I would  rather reduce this whole condition to directly exit based on the `non-zero` exit code of the response for the `describe` call.

Ganiredi: @hakuna-matatah  If i understand correct you want me to display more like this instead of just `does not exist` response right ?

```
if [ $? -eq 0 ]; then
  printf "Stack exists\n"
  exit 0
else
  printf "Stack does not exist\n"
  exit 1

File: tests/tasks/teardown/awscli-vpc-delete.yaml [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+spec:
+   description: |
+    This Task can be used to delete CloudFormation stack containing VPC resources that was used for EKS clusters.
+   params:
+    - name: stack-name
+      description: The name of the VPC name you want to delete.
+    - name: region
+      default: "us-west-2"
+   steps:
+      - name: awscli-delete-vpc
```
mengqiy, line 15: Formatting nit. Please fix indentation. You should remove one space.

tests/tasks/teardown/awscli-vpc-delete.yaml
mengqiy: +1
Also do we actually want to exit when the stack is not found. Exit with non-0 will cause pipeline to stop.

mengqiy: What harish mean is to check the error code from `aws cloudformation describe-stacks`. Checking error message is a little fragile.

hakuna-matatah: >Also do we actually want to exit when the stack is not found. Exit with non-0 will cause pipeline to stop.

@mengqiy I think we should exit with non zero when stack not found as  it's very unexpected to have vpc stack disappear if every other task in test pipeline run was successful, we want to understand if those cases arise and not mask them. wdyt ?

.github/workflows/main.yml
```diff
@@ -30,3 +30,10 @@ jobs:
       run: |
         GOOS=${{ matrix.os }} GOARCH=${{ matrix.arch }} go build -o ./bin/kitctl ./cmd/kitctl/
       working-directory: substrate
+    - uses: docker/login-action@v1
```
hakuna-matatah, line 33: why do we need this for this PR ? Wrong commit got pushed to this PR. Please correct it.

Ganiredi: Testing this in my fork repo ,Unfortunately it seems was added to previous PR's ,Please neglect it.Thanks

tests/tasks/teardown/awscli-vpc-delete.yaml
```diff
@@ -0,0 +1,34 @@
...
+      image: alpine/k8s:1.22.6
+      script: |
+        #!/bin/bash
+        echo "Approving KCM requests"
+        kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
+        aws sts get-caller-identity 
+        # Check if the stack exists
+        aws cloudformation --region $(params.region) describe-stacks --stack-name $(params.stack-name)
+        if [ $? -ne 0 ]; then
+        echo "Stack $(params.stack-name) not found ,Exiting..."
```
mengqiy, line 25: Syntax nit: "Stack $(params.stack-name) not found. Exiting..."

Please add 2 spaces to readability in the if else statement.

mengqiy (approved):

tests/tasks/teardown/awscli-vpc-delete.yaml
```diff
@@ -0,0 +1,34 @@
...
+        #!/bin/bash
+        echo "Approving KCM requests"
+        kubectl certificate approve $(kubectl get csr | grep "Pending" | awk '{print $1}')  2>/dev/null || true
+        aws sts get-caller-identity 
+        # Check if the stack exists
+        aws cloudformation --region $(params.region) describe-stacks --stack-name $(params.stack-name)
+        if [ $? -ne 0 ]; then
+          echo "Stack $(params.stack-name) not found. Exiting..."
+          exit 1
+        else
```
hakuna-matatah, line 27: You don't really need this else block here.

Ganiredi: Thought it would be more beneficial for users if the process of deleting a stack was more clearly communicated and made more visible .

PR #355, opened 2023-01-26
Adds Prometheus Write Permission to IAM Role

ashishranjan738:
This commit adds Prometheus Write Permission to IAM Role for EC2 nodes, so that Prometheus can write metrics to AMP.


hakuna-matatah (approved):
LGTM

PR #357, opened 2023-02-02
Forcing the use of IMDSV2 endpoint through lauch template

adityavenneti:
Description of changes:
Disabling IMDSV1 by forcing the use of IMDSV2 through launch template


hakuna-matatah (approved):
lgtm

PR #360, opened 2023-02-06
Feature to delete cluster resources when TTL expires

prateekgogia:
Issue #, if available:

Description of changes:
When a TTL is set for a CP object, operator will clean up the resources. This is to avoid leaking CP objects in CI when CP objects are not being deleted and they tend to run for too long.

### DIFF
--- a/operator/pkg/controllers/controller.go
+++ b/operator/pkg/controllers/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/awslabs/kubernetes-iteration-toolkit/operator/pkg/apis/controlplane/v1alpha1"
 	"github.com/awslabs/kubernetes-iteration-toolkit/operator/pkg/errors"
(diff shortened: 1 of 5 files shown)


ashishranjan738:
Nice !!

operator/pkg/controllers/controller.go
```diff
@@ -106,3 +119,34 @@ func (c *GenericController) reconcile(ctx context.Context, resource Object, pers
...
+	if !ok {
+		return false
+	}
+	if cp.Spec.TTL != "" {
+		duration, err := time.ParseDuration(cp.Spec.TTL)
+		if err != nil {
+			zap.S().Errorf("parsing TTL duration, %w", err)
+			return false
+		}
+		deleteAfter := resource.GetCreationTimestamp().Add(duration)
```
ashishranjan738, line 145: minor: i think this logic can be reused to check ttls for other components too..like data plane. However not a blocker.

ashishranjan738 (approved):

operator/pkg/controllers/controller.go
```diff
@@ -106,3 +119,34 @@ func (c *GenericController) reconcile(ctx context.Context, resource Object, pers
...
+	zap.S().Infof("[%s] Successfully deleted resources", resource.GetName())
+	return result, nil
+}
+
+func ttlExpiredForControlPlane(resource Object) bool {
+	cp, ok := resource.(*v1alpha1.ControlPlane)
+	if !ok {
+		return false
+	}
+	if cp.Spec.TTL != "" {
```
hakuna-matatah, line 139: I was wondering can we have some sane defaults ? 
I'm not sure if we can expect customers set this value by default and we might end up with these resources ?

prateekgogia: Setting default value is tough since we will change the existing functionality and it might lead to surprises if this field is not set in the CP spec. Some use cases like CI  testing require 24 hr TTL other like performance test might require like 7 days retention. Open to ideas.

hakuna-matatah: Hmmm, I understand. Given it's in alpha/beta phase, may be we can take a stance now ?
For example - IIRC, default provisioners of karpenter has a very aggressive TTL set today.
Given idea of these guest clusters is that, it is ephemeral by its nature. We can be conservative and start with 10days and document in README ? wdyt ?

mengqiy: Ideally we should extend it for some period of time when the cluster is not "idle". e.g. CPU > 20%. But this makes it harder to implement.

I guess we can add a default for TTL, given we are still pre-GA.
This can potentially lead to surprises, we should add it to release notes and README.

mengqiy:
How soon can we detect an expired CP?

operator/pkg/controllers/controller.go
prateekgogia: > This can potentially lead to surprises, we should add it to release notes and README.

Scenario I am worried about is, lets say someone has a created guest cluster in the past and left it running for few days, cleaned up manually. With default TTL now the cluster will get cleaned up automatically which can be a surprise for someone say StormForge for example.

prateekgogia: Also can we add the default behavior in another PR? Trying to get the 1.25 cluster creation rolled out as well in this PR

PR #365, opened 2023-02-09
Github workflow that uses kitctl to bootstrap a test environment

Ganiredi:
Issue #, if available:

Description of changes:

### DIFF
--- a/.github/actions/install-go-and-dependencies/action.yml
+++ b/.github/actions/install-go-and-dependencies/action.yml
@@ -1,0 +1,18 @@
+name: setup-go-and-cache
+description: 'Set up go and cache'
+runs:
+  using: "composite"
+  steps:
+  - uses: actions/checkout@v3
+  - uses: actions/cache@v3
+    with:
+        path: |
+          ~/.cache/go-build
+          ~/go/pkg/mod
+          ~/go/bin/
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+  - uses: actions/setup-go@v3
+    with:
+        go-version: '1.17.1'  
--- a/.github/workflows/kitctl-test.yml
+++ b/.github/workflows/kitctl-test.yml
@@ -1,0 +1,218 @@
+name: This workflow automates the build, testing, and bootstrapping of a test environment on Kubernetes using KIT CLI
+on:
+  push:
+    branches: [main]
+permissions:
+  id-token: write
+  pull-requests: write
+  contents: read
+jobs:
+  kit:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest,macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/install-go-and-dependencies
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::638256630554:role/kit-github-workflow
+          role-session-name: githubworkflowsession
+          aws-region: us-west-2
+      - name: Install kubectl
+        run: |
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/darwin/amd64/kubectl"  
+          fi
+          chmod +x kubectl
+          sudo mv kubectl /usr/local/bin/
+      - name: Build kitctl, provision Kit Environment and KIT guest cluster.
+        run: |
+          allVersions=("1-21" "1-22" "1-23" "1-24" "1-25")
+          cd ${HOME}/work/kubernetes-iteration-toolkit/kubernetes-iteration-toolkit/substrate/cmd/kitctl
+          go build  .
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            os_type="linux"    
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+          fi 
+          ./kitctl bootstrap kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+          kubectl get pods -A
+          for version in "${allVersions[@]}"; do
+            K8S_VERSION=${version//-/.}
+            GUEST_CLUSTER_NAME="guest-${os_type}-${version}-${{ github.run_id }}-${{github.run_attempt}}"
+            cat <<-EOF | kubectl apply -f -
+              apiVersion: kit.k8s.sh/v1alpha1
+              kind: ControlPlane
+              metadata:
+                name: ${GUEST_CLUSTER_NAME} # Desired Cluster name
+              spec:
+                kubernetesVersion: "${K8S_VERSION}"
+                etcd:
+                  replicas: 3
+                master:
+                  apiServer:
+                    replicas: 1  
+          EOF
+            for (( time=0; time<=600; time+=5 )); do
+              controlplane_object=$(kubectl get controlplane ${GUEST_CLUSTER_NAME}  -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
+              if [[ "$controlplane_object" == "True" ]]; then
+                echo "Control plane objects are ready"
+                break;
+              else
+                echo "Control plane objects are not ready"
+              fi
+              if [[ "$time" -eq 590 ]]; then
+                echo "Timeout: Control plane objects are not ready after 10 minutes"
+                kubectl logs deploy/kit-controller -n kit
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl wait --for=condition=Ready pods  --all --timeout=5m
+            GUESTKUBECONFIG=/tmp/kubeconfig
+            kubectl get secret ${GUEST_CLUSTER_NAME}-kube-admin-config -ojsonpath='{.data.config}' | base64 -d >$GUESTKUBECONFIG
+            cat <<-EOF | kubectl apply -f -
+              apiVersion: kit.k8s.sh/v1alpha1
+              kind: DataPlane
+              metadata:
+                name: ${GUEST_CLUSTER_NAME}-nodes
+              spec:
+                clusterName: ${GUEST_CLUSTER_NAME} # Associated Cluster name
+                nodeCount: 2
+          EOF
+            for (( time=0; time<=600; time+=5 )); do
+              dataplane_object=$(kubectl get dataplane ${GUEST_CLUSTER_NAME}-nodes  -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
+              if [[ "$dataplane_object" == "True" ]]; then
+                echo "Dataplane objects are ready"
+                break;
+              else
+                echo "Dataplane objects are not ready"
+              fi
+              if [[ "$time" -eq 590 ]]; then
+                echo "Timeout: Dataplane objects are not ready after 10 minutes"
+                kubectl logs deploy/kit-controller -n kit 
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl --kubeconfig=$GUESTKUBECONFIG apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.10/config/master/aws-k8s-cni.yaml
+            kubectl --kubeconfig=$GUESTKUBECONFIG rollout status daemonset/aws-node -n kube-system --timeout 5m
+            for (( time=0; time<=220; time+=5 )); do
+              ready_node=$(kubectl --kubeconfig=$GUESTKUBECONFIG get nodes 2>/dev/null | grep -w Ready | wc -l)
+              if [[ "$ready_node" -eq 2 ]]; then
+                echo "Dataplane nodes are ready"
+                break;
+              else
+                echo "Dataplane nodes are not ready"
+              fi
+              if [[ "$time" -eq 210 ]]; then
+                echo "Timeout: Dataplane nodes are not ready after 4 minutes"
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl --kubeconfig=$GUESTKUBECONFIG create deployment nginx-deployment --image=nginx
+            kubectl --kubeconfig=$GUESTKUBECONFIG rollout status deployment/nginx-deployment --timeout 5m
+            kubectl --kubeconfig=$GUESTKUBECONFIG expose deployment nginx-deployment --name=nginx-service --port=80 --target-port=80 --type=ClusterIP
+            kubectl --kubeconfig=$GUESTKUBECONFIG certificate approve $(kubectl --kubeconfig=$GUESTKUBECONFIG get csr | grep Pending | awk '{print $1}')
+            kubectl --kubeconfig=$GUESTKUBECONFIG run -i --rm --restart=Never curly-pod  --image=curlimages/curl --command -- curl -s 'http://nginx-service:80'
+          done
+        working-directory: substrate
+      - name:  Delete the Guest Control plane and Dataplane
+        if: always()
+        run: |
+          allVersions=("1-21" "1-22" "1-23" "1-24" "1-25")
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            os_type="linux"    
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+          fi 
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+          kubectl config current-context
+          for version in "${allVersions[@]}"; do
+            GUEST_CLUSTER_NAME="guest-${os_type}-${version}-${{ github.run_id }}-${{github.run_attempt}}"
+            if kubectl get dataplane  $GUEST_CLUSTER_NAME-nodes > /dev/null 2>&1; then
+              kubectl delete dataplane  $GUEST_CLUSTER_NAME-nodes
+              echo "dataplane deleted."
+            else
+              echo "No dataplane found  $GUEST_CLUSTER_NAME-nodes."
+            fi
+            if kubectl get controlplane  $GUEST_CLUSTER_NAME > /dev/null 2>&1; then
+              kubectl delete controlplane  $GUEST_CLUSTER_NAME
+              echo "Controlplane deleted."
+            else
+              echo "No controlplane found  $GUEST_CLUSTER_NAME."
+            fi
+          done 
+          #Adding sleep to ensure that the resources are deleted like EG:Load balancer.
+          sleep 120
+        continue-on-error: true
+        working-directory: substrate
+      - name:  Delete the kit environment
+        if: always()
+        run: |
+          #During the first attempt, "kitctl" was unable to delete some resources sporadically. To ensure that enough time is given for the resources to be deleted, wait time is added around 280 seconds before attempting to delete the kit again.
+          #and for the second iteration with a wait time of 30 seconds to ensure all resources are successfully deleted as we are left with few resources sometimes.
+          #TODO: we should fix `kitctl delete` to make it more robust and not hang forever.
+          cd ${HOME}/work/kubernetes-iteration-toolkit/kubernetes-iteration-toolkit/substrate/cmd/kitctl
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            os_type="linux"    
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+            # installing coreutils for macos to use gtimeout as timeout is not available in macos by default,as kitctl hangs sparodically 
+            /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+            brew install coreutils
+          fi 
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+          # Set the maximum number of attempts
+          MAX_ATTEMPTS=2
+          # Set the delay between retries in seconds
+          RETRY_DELAY=30
+          # Set the maximum duration in seconds
+          MAX_DURATION=280
+          # Set the start time
+          START_TIME=$(date +%s)
+          # Set a flag to indicate whether the command was successfully completed or not
+          COMMAND_COMPLETED=false
+          # Attempt to delete the resource multiple times
+          for i in $(seq 1 $MAX_ATTEMPTS); do
+            # Run the command with a timeout
+            if [[ "${{ matrix.os }}" == "macos-latest"  ]]; then
+              gtimeout $MAX_DURATION ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+            else
+              timeout $MAX_DURATION ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+            fi
+            # Check if the command timed out
+            status=$?
+            if [ $status -eq 0 ]; then
+              echo "Command executed successfully for $i iteration."
+              COMMAND_COMPLETED=true
+              break
+            elif [ $status -eq 124 ]; then
+              echo "Command timed out. Retrying in $RETRY_DELAY seconds."
+            elif [ $status -eq 143 ]; then
+              echo "Command terminated by SIGTERM signal. Retrying in $RETRY_DELAY seconds."
+            else
+              echo "Command failed with exit code $status."
+            fi
+            # Wait for the specified delay before trying again
+            sleep $RETRY_DELAY
+            # Calculate the elapsed time
+            ELAPSED_TIME=$(( $(date +%s) - $START_TIME ))
+            # If the elapsed time exceeds the maximum duration, print a message and exit the loop
+            if [ $ELAPSED_TIME -gt $MAX_DURATION ]; then
+              echo "Maximum duration exceeded. Exiting the loop."
+              break
+            fi
+          done
+          # Check if the command was successfully completed
+          if [ "$COMMAND_COMPLETED" = false ]; then
+            echo "Failed to delete resource after $MAX_ATTEMPTS attempts."
+          fi
+        continue-on-error: true
+        working-directory: substrate
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![kitctl](https://github.com/awslabs/kubernetes-iteration-toolkit/actions/workflows/kitctl-test.yml/badge.svg)](https://github.com/awslabs/kubernetes-iteration-toolkit/actions/workflows/kitctl-test.yml) **`kitctl`** test coverage when new modifications are committed to this repository
+
 # What is Kubernetes Iteration Toolkit?
 
 ## What is KIT?


Ganiredi:
Unfortunately,I deleted the forked repo as I had do many changes that points to different repos to test which made almost 350+ commits a head of the main repo  which made me feel that my repo bit clumsy,Here I opened a new pull request and closed the previous pull request https://github.com/awslabs/kubernetes-iteration-toolkit/pull/353,as I cannot use once i deleted the fork repo and open new one.

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
+name: InstallGo and configure aws creds
```
mengqiy, line 1: We are not reusing this. Why don't we embed it?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,60 @@
...
+  pull-requests: write
+  contents: read
+jobs:
+  kit:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
```
mengqiy, line 17: Why do we checkout twice? There is another checkout in `install-go-and-dependencies`?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: ./.github/actions/install-go-and-dependencies
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: arn:aws:iam::953421922360:role/kit-Github-workflow
+          role-session-name: githubworkflowsession
+          aws-region: us-west-2
+      - name: Install kubectl
```
mengqiy, line 24: I don't find we are using kubectl anywhere.

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
...
+description: 'Installs Go Downloads and configure aws credentials '
```
mengqiy, line 2: This job doesn't configure credentials

.github/workflows/trigger-kit.yml
```diff
@@ -0,0 +1,13 @@
+name: Trigger-kit
```
mengqiy, line 1: maybe call it `trigger-kit-CI`

.github/workflows/trigger-kit.yml
```diff
@@ -0,0 +1,13 @@
...
+    branches: [main]
+  push:
+    branches: [main]
+jobs:
+  depend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Message
+        run: echo "Triggers second workflow kit-bootstrap"
```
mengqiy, line 13: echo "Triggers workflow kitctl-CI"

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,60 @@
...
+          fi
+          chmod +x kubectl
+          sudo mv kubectl /usr/local/bin/
+      - name: Builds the KIT CLI ,Provisions the Kit Environment and deletes it
+        run: |
+          cd ${HOME}/work/kubernetes-iteration-toolkit/kubernetes-iteration-toolkit/substrate/cmd/kitctl
+          go build -v .
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+              ./kitctl bootstrap kit-github-linux-${{ github.run_id }}
+              export KUBECONFIG=${HOME}/.kit/env/kit-github-linux-${{ github.run_id }}/etc/kubernetes/admin.conf
```
mengqiy, line 39: We should check the management cluster is working by using kubectl

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,60 @@
...
+              for i in {1..5}; do
+                  ./kitctl delete kit-github-linux-${{ github.run_id }}
+                  if [ $i -eq 1 ]; then
+                     sleep 220
+                  else
+                    sleep 60
+                  fi
+               done
+           else
+              ./kitctl bootstrap kit-github-darwin-${{ github.run_id }}
```
mengqiy, line 49: No need to repeat yourself.
We can embed `{{ matrix.os }}` in the name.

.github/workflows/trigger-kit.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+jobs:
+  depend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
mengqiy, line 11: Checkout doesn't do anything here.

mengqiy:
Please update the PR description as we discussed before to include the context about why we take this approach and why alternatives don't work.

mengqiy:
Also please squash commits. No need to have 18 commits.

.github/actions/install-go-and-dependencies/action.yml
Ganiredi: Test coverage for kit guest cluster and Eks guest cluster also needs installation of GO,Instead of re writing the installation code again we can reuse it.

.github/workflows/kit-bootstrap.yml
Ganiredi: I will add `kubectl get pods -A` to do a small test

.github/actions/install-go-and-dependencies/action.yml
Ganiredi: Ack

.github/workflows/trigger-kit.yml
Ganiredi: Ack

.github/workflows/trigger-kit.yml
Ganiredi: Ack

.github/workflows/kit-bootstrap.yml
Ganiredi: To create the roles in `IAM` it has to satisfy constraint: Member must have length less than or equal to 64.By embedding the `{{ matrix.os }}`  will create larger than 64 characters which fails to bootstrap the kit-environment.

.github/workflows/kit-bootstrap.yml
Ganiredi: Its Necessary to checkout twice ,Otherwise `Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/kubernetes-iteration-toolkit/kubernetes-iteration-toolkit/.github/actions/install-go-and-dependencies'. Did you forget to run actions/checkout before running your local action?` throws an error ,Tested locally with and without `- uses: actions/checkout@v2`

.github/workflows/trigger-kit.yml
Ganiredi: Ack

Ganiredi:
> 

In order to use AWS resources in GitHub Action workflow there are some limitations ,When we a Create a PR from the fork to the upstream repository workflow wont works because GitHub Secrets are not passed to the runner when a workflow is triggered from a forked repository, In order to overcome to this @mengqiy and I came up with an workaround to use a secondary GitHub Action Workflow that is triggered on [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run). Here first workflow action will be triggered and basing on that the second GA will authenticate properly(AWS creds). But there are certain limitations when we are triggering the workflow_run events ,Where the second workflow that was triggered by the first workflow will always run in the background and doesn't have visibility in checks.

References:
(1)[https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-featu[…]-repository/managing-github-actions-settings-for-a-repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories)

(2)https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
+name: InstallGo 
+description: 'Installs Go Downloads'
```
mengqiy, line 2: `Set up go and cache`

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
+name: InstallGo
```
mengqiy, line 1: `setup-go-and-cache`

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
+name: InstallGo 
+description: 'Installs Go Downloads'
+runs:
+  using: "composite"
+  steps:
```
mengqiy, line 5: You should audit existing workflow files and replace setup-go-and-cache with yours. e.g. https://github.com/awslabs/kubernetes-iteration-toolkit/blob/06baa4e9112320d2dcbc6c25e48ac75a1b6151cd/.github/workflows/build.yaml#L13-L24

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  kit:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/install-go-and-dependencies
+      - uses: aws-actions/configure-aws-credentials@v1
```
mengqiy, line 19: I'd expect you to reference things like `AWS_ACCESS_KEY_ID_PERFLAB`. Why are you not using them now?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,53 @@
...
+          go build -v .
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            os_type="linux"
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+          fi 
+          ./kitctl bootstrap kit-github-${os_type}-${{ github.run_id }}
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}/etc/kubernetes/admin.conf
+          kubectl get pods -A
+          for i in {1..5}; do
```
mengqiy, line 45: Add comments about why retry here.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,53 @@
...
+            ./kitctl delete kit-github-${os_type}-${{ github.run_id }}
+            if [ $i -eq 1 ]; then
+              sleep 220
```
mengqiy, line 48: Why sleep different amount of time?
If it's a must-have, we should have a comment here.

.github/actions/install-go-and-dependencies/action.yml
```diff
@@ -0,0 +1,18 @@
+name: InstallGo 
+description: 'Installs Go Downloads'
+runs:
+  using: "composite"
```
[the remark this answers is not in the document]
Ganiredi: Using a composite action can make workflow code more modular and easier to maintain. Instead of repeating the same set of steps across multiple workflows, we can define a composite action once and then we reference it wherever it's needed,Here the i used this action for setup go and used as a single line ,I will use the same action for the next workflows that are used to do test coverage for kit guest cluster and Eks guest cluster.

.github/workflows/kit-bootstrap.yml
Ganiredi: while AWS access keys can be a valid way to authenticate and authorize access to resources, OIDC provides a more secure, scalable, and manageable solution.

.github/actions/install-go-and-dependencies/action.yml
Ganiredi: updating with latest versions.

Ganiredi:
The below workflow runs on both `Ubuntu-latest` and `macOS`, however, the `macOS` run takes a significantly longer duration compared to the `Ubuntu-latest` run. This is because, by the time the `Ubuntu-latest` run completes the creation of the guest KIT cluster from 1-21 to 1-24, the macOS run is still struggling to create the control plane guest nodes and is experiencing a lot of network latency issues it seems. This results in the entire workflow being impacted and slowed down. To address this, we need to find a way to optimize the network latency for the macOS run or consider using an alternative solution.

[macos-latest]->https://github.com/Ganiredi/kubernetes-iteration-toolkit/actions/runs/4266143419/jobs/7426309435

[ubuntu-latest]->https://github.com/Ganiredi/kubernetes-iteration-toolkit/actions/runs/4266000172/jobs/7425999484

[macos-latest][ubuntu-latest]->https://github.com/Ganiredi/kubernetes-iteration-toolkit/actions/runs/4264715980

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+            os_type="linux"    
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+          fi 
+          ./kitctl bootstrap kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+          kubectl get pods -A
+          for i in 21 22 23 24 25; do
+            GUEST_CLUSTER_NAME="guest-${os_type}-1-${i}-${{ github.run_id }}-${{github.run_attempt}}"
+            K8S_VERSION=$(echo ${GUEST_CLUSTER_NAME} | sed 's/guest-linux-1-//;s/-.*//;s/^/1./')
```
mengqiy, line 47: I don't understand what you are doing here. How can you get `K8S_VERSION` from `GUEST_CLUSTER_NAME `?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+                name: ${GUEST_CLUSTER_NAME} # Desired Cluster name
+              spec:
+                kubernetesVersion: "${K8S_VERSION}"
+                etcd:
+                  replicas: 3
+                master:
+                  apiServer:
+                    replicas: 1  
+          EOF
+           for (( time=0; time<=1000; time+=5 )); do
```
mengqiy, line 61: 600 seconds should be enough.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+              controlplane_object=$(kubectl get controlplane ${GUEST_CLUSTER_NAME}  -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
+              if [[ "$controlplane_object" == "True" ]]; then
+                echo "Control plane objects are ready"
+                break;
+              else
+                echo "Control plane objects are not ready"
+              fi
+              if [[ "$time" -eq 990 ]]; then
+                echo "Timeout: Control plane objects are not ready after 20 minutes"
```
mengqiy, line 70: We should use 10 minutes timeout as we discussed offline

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+                kubectl logs deploy/kit-controller -n kit
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl wait --for=condition=Ready pods  --all --timeout=5m
+            kubectl get secret ${GUEST_CLUSTER_NAME}-kube-admin-config -ojsonpath='{.data.config}' | base64 -d > /tmp/kubeconfig
```
mengqiy, line 77: `/tmp/kubeconfig` has been used many times below. Maybe consider using a variable.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+            kubectl --kubeconfig=/tmp/kubeconfig apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.10/config/master/aws-k8s-cni.yaml
+            kubectl --kubeconfig=/tmp/kubeconfig rollout status daemonset/aws-node -n kube-system
```
mengqiy, line 79: This is not doing anything useful when there are no worker nodes.
I'd suggest to change it to:
- Create DP object
- Check DP reporting ready in staus
- Apply CNI
- Use `rollout status` to wait for completion.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+              if [[ "$time" -eq 290 ]]; then
+                echo "Timeout: Dataplane nodes are not ready after 10 minutes"
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl --kubeconfig=/tmp/kubeconfig create deployment nginx-deployment --image=nginx
+            kubectl --kubeconfig=/tmp/kubeconfig rollout status deployment/nginx-deployment
+            kubectl --kubeconfig=/tmp/kubeconfig expose deployment nginx-deployment --name=nginx-service --port=80 --target-port=80 --type=ClusterIP
+            kubectl --kubeconfig=/tmp/kubeconfig certificate approve $(kubectl --kubeconfig=/tmp/kubeconfig get csr | grep Pending | awk '{print $1}')
```
mengqiy, line 121: This is only useful for local debugging. But it's not useful to run in this test.
What are you trying to do with it?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+            kubectl --kubeconfig=/tmp/kubeconfig run -i --rm --restart=Never curly-pod  --image=curlimages/curl --command -- curl -s 'http://nginx-service:80'
```
mengqiy, line 122: If curl fails, do we exit here?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+            kubectl delete controlplane ${GUEST_CLUSTER_NAME}
+          done
+          #During the first attempt, "kitctl" was unable to delete some resources sporadically. To ensure that enough time is given for the resources to be deleted, a longer wait time of 220 seconds is added before attempting to delete the kit again.
+          #and for the remaining 4 iterations with a wait time of 60 seconds to ensure all resources are successfully deleted as we are left with few resources sometimes.
+          for i in {1..5}; do
+            ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
```
mengqiy, line 128: Deleting subtrate should happen in a separate step so that no resources are leaked even if early steps fail.
Ref: https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f

.github/actions/install-go-and-dependencies/action.yml
mengqiy: What do you mean?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,135 @@
...
+              controlplane_object=$(kubectl get controlplane ${GUEST_CLUSTER_NAME}  -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null)
+              if [[ "$controlplane_object" == "True" ]]; then
+                echo "Control plane objects are ready"
+                break;
+              else
+                echo "Control plane objects are not ready"
+              fi
+              if [[ "$time" -eq 990 ]]; then
+                echo "Timeout: Control plane objects are not ready after 20 minutes"
+                kubectl logs deploy/kit-controller -n kit
```
mengqiy, line 71: I'd suggest also logging the pods status: `kubectl get pods -o wide`.

.github/actions/install-go-and-dependencies/action.yml
Ganiredi: This action.yml we can use it for later workflows for provisioning the eks guest cluster and also we can update the existing ones in our workflows

.github/workflows/kit-bootstrap.yml
Ganiredi: Example 
GUEST_CLUSTER_NAME=guest-linux-1-25-2846287626-8
K8S_VERSION=$(echo ${GUEST_CLUSTER_NAME} | sed 's/guest-linux-1-//;s/-.*//;s/^/1./')
Here we want the k8 version only in  string type first it removes the prefix "guest-linux-1-" ,removes the suffix after the first dash "-" and adds the prefix "1." to the extracted version number

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
...
+              if [[ "$time" -eq 150 ]]; then
+                echo "Timeout: Dataplane nodes are not ready after 3 minutes"
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl kubeconfig_guestcluster create deployment nginx-deployment --image=nginx
+            kubectl kubeconfig_guestcluster rollout status deployment/nginx-deployment
+            kubectl kubeconfig_guestcluster expose deployment nginx-deployment --name=nginx-service --port=80 --target-port=80 --type=ClusterIP
+            kubectl kubeconfig_guestcluster certificate approve $(kubectl kubeconfig_guestcluster get csr | grep Pending | awk '{print $1}')
```
mengqiy, line 123: Echo what I asked in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/365#discussion_r1117791130
Is this doing anything useful for us?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
...
+              kubectl config current-context
+              for i in {1..4}; do
+                ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+                if [ $i -eq 1 ]; then
+                  sleep 220
+                else
+                  sleep 60
+                fi
+              done
+      - name: Fail the workflow
```
mengqiy, line 149: Why do we need this?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
...
+                echo "Dataplane objects are not ready"
+              fi
+              if [[ "$time" -eq 590 ]]; then
+                echo "Timeout: Dataplane objects are not ready after 10 minutes"
+                kubectl logs deploy/kit-controller -n kit 
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl kubeconfig_guestcluster apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.10/config/master/aws-k8s-cni.yaml
```
mengqiy, line 104: > `kubeconfig_guestcluster`

You need to make it a bash variable

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
+name: This workflow automates the build, testing, and bootstrapping of a test environment on Kubernetes using KIT CLI
+on:
```
mengqiy, line 2: After examining it, I realized that making it triggered by another workflow can only exercise code lives in the main branch, but can't test against new changes from PR.
It makes more sense to make this test a post-submit test.
It's simple to achieve that, just make the `on` section to below.
```
on:
  push:
    branches: [main]
permissions:
  ...
```

We can add a GH action badge (https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) in README once we verify everything works fine.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
...
+          go build  .
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            os_type="linux"    
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            os_type="darwin"
+          fi 
+          ./kitctl bootstrap kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+          export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+          kubectl get pods -A
+          for i in 21 22 23 24 25; do
```
mengqiy, line 46: I believe I have talked about this offline.
I'd suggest defining an array like `allVersions=("1.21" "1.22" "1.23" "1.24" "1.25")` at the very top of this script so that it's more visible.
Then you can have something like `for version in ${allVersions[@]}`.
This help a lot with code readability.

.github/workflows/kit-bootstrap.yml
mengqiy: Why make it so complicated when `${i}` holds the information for you.
```
K8S_VERSION=1-${i}
```

.github/workflows/trigger-kit.yml
```diff
@@ -0,0 +1,12 @@
+name: trigger-kit-CI
```
mengqiy, line 1: We will no longer need this file once we change it to post-submit

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,152 @@
...
+    types: [completed]
+permissions:
+  id-token: write
+  pull-requests: write
+  contents: read
+jobs:
+  kit:
+    strategy:
+      matrix:
+        os: [ macos-latest ]
```
mengqiy, line 14: We want to test on both linux and macos.

.github/workflows/kit-bootstrap.yml
mengqiy: You never define `kubeconfig_guestcluster` anywhere. I don't think this script will work. Have you tested it?

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,154 @@
...
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl  --kubeconfig=/tmp/kubeconfig create deployment nginx-deployment --image=nginx
+            kubectl  --kubeconfig=/tmp/kubeconfig rollout status deployment/nginx-deployment
+            kubectl  --kubeconfig=/tmp/kubeconfig expose deployment nginx-deployment --name=nginx-service --port=80 --target-port=80 --type=ClusterIP
+            kubectl  --kubeconfig=/tmp/kubeconfig certificate approve $(kubectl kubeconfig_guestcluster get csr | grep Pending | awk '{print $1}')
+            kubectl  --kubeconfig=/tmp/kubeconfig run -i --rm --restart=Never curly-pod  --image=curlimages/curl --command -- curl -s 'http://nginx-service:80'
+            kubectl delete controlplane ${GUEST_CLUSTER_NAME}
```
mengqiy, line 127: This should be moved to the cleanup step.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,154 @@
...
+          done
+        working-directory: substrate
+      - name:  Delete the kit environment 
+        if: always()
+        run: |
```
mengqiy, line 132: We should add continue-on-error for this step.

.github/workflows/kit-new.yml
```diff
@@ -0,0 +1,165 @@
...
+               if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+                os_type="linux"    
+              elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+                os_type="darwin"
+              fi 
+              export KUBECONFIG=${HOME}/.kit/env/kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}/etc/kubernetes/admin.conf
+              kubectl config current-context
+              for version in "${allVersions[@]}"; do
+                GUEST_CLUSTER_NAME="guest-${os_type}-${versions}-${{ github.run_id }}-${{github.run_attempt}}"
+                kubectl delete controlplane ${GUEST_CLUSTER_NAME}
```
mengqiy, line 151: I'd suggest splitting CP cleanup from kit env cleanup, since we want to ensure we execute both cleanup no matter if one of them fails.
Currently if `kubectl delete controlplane`, the cleanup for kit will not be executed.

mengqiy (commented):
There is one comment (https://github.com/awslabs/kubernetes-iteration-toolkit/pull/365#discussion_r1122170622) that has not been addressed. Otherwise LGTM

mengqiy (approved):

hakuna-matatah:
Let's make sure we run this in dedicated Team account for this workflow as mentioned offline in the past. Making sure, we don't  have single point of failure when you are not available.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,213 @@
...
+        run: | 
+              #During the first attempt, "kitctl" was unable to delete some resources sporadically. To ensure that enough time is given for the resources to be deleted, a longer wait time of 220 seconds is added before attempting to delete the kit again.
+              #and for the remaining 4 iterations with a wait time of 30 seconds to ensure all resources are successfully deleted as we are left with few resources sometimes.
+              cd ${HOME}/work/kubernetes-iteration-toolkit/kubernetes-iteration-toolkit/substrate/cmd/kitctl
+              if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+                os_type="linux"    
+              elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+                os_type="darwin"
+                /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+                brew install coreutils
```
mengqiy, line 167: Why do we need this?
When you have something that's not obvious, you should add a comment.

.github/workflows/kit-bootstrap.yml
```diff
@@ -0,0 +1,213 @@
...
+              MAX_ATTEMPTS=5
+              # Set the delay between retries in seconds
+              RETRY_DELAY=30
+              # Set the maximum duration in seconds
+              MAX_DURATION=220
+              # Set the start time
+              START_TIME=$(date +%s)
+              # Set a flag to indicate whether the command was successfully completed or not
+              COMMAND_COMPLETED=false
+              # Attempt to delete the resource multiple times
```
mengqiy, line 180: I'm OK with retrying deleting in the short term, but please add a TODO saying we should fix `kitctl delete` to make it more robust and not hang forever.

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,216 @@
...
+      - name: Install kubectl
+        run: |
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/darwin/amd64/kubectl"  
+          fi
+          chmod +x kubectl
+          sudo mv kubectl /usr/local/bin/
+      - name: Builds the KIT CLI ,Provisions the Kit Environment and KIT guest cluster.
```
mengqiy, line 32: Grammar nit and also trying to make it more concise
Build kitctl, provision Kit Environment and KIT guest cluster.

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,216 @@
...
+              fi
+              if [[ "$time" -eq 590 ]]; then
+                echo "Timeout: Dataplane objects are not ready after 10 minutes"
+                kubectl logs deploy/kit-controller -n kit 
+                exit 1
+              fi
+              sleep 5
+            done
+            kubectl --kubeconfig=$GUESTKUBECONFIG apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.10/config/master/aws-k8s-cni.yaml
+            kubectl --kubeconfig=$GUESTKUBECONFIG rollout status daemonset/aws-node -n kube-system
```
mengqiy, line 105: We should specify the --timeout flag in case something wrong happen.
Currently it may stuck forever here until GH action timeout.

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,216 @@
...
+          if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
+          elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
+            curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -L -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/darwin/amd64/kubectl"  
+          fi
+          chmod +x kubectl
+          sudo mv kubectl /usr/local/bin/
+      - name: Builds the KIT CLI ,Provisions the Kit Environment and KIT guest cluster.
+        run: |
+          allVersions=("1.21" "1.22" "1.23" "1.24" "1.25")
```
mengqiy, line 34: You use `allVersions=("1-21" "1-22" "1-23" "1-24" "1-25")` below. You should be consistent.

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,216 @@
...
+              timeout $MAX_DURATION ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+            fi
+            # Check if the command timed out
+            status=$?
+            if [ $status -eq 124 ]; then
+              echo "Command timed out. Retrying in $RETRY_DELAY seconds."
+            elif [ $status -eq 143 ]; then
+              echo "Command terminated by SIGTERM signal."  
+            else
+              echo "Command timed out  successfully for $i iteration."
```
mengqiy, line 198: > Command timed out successfully

This is confusing. If it times out, it should be a failure.
Is it a success if exit code is 0?

mengqiy (commented):
A few minor comments

Ganiredi:
Eg Workflow run: https://github.com/Ganiredi/kubernetes-iteration-toolkit/actions/runs/4863420273/workflow

.github/workflows/kitctl-test.yml
mengqiy: This is not addressed.

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,215 @@
...
+              timeout $MAX_DURATION ./kitctl delete kit-github-${os_type}-${{ github.run_id }}-${{github.run_attempt}}
+            fi
+            # Check if the command timed out
+            status=$?
+            if [ $status -eq 124 ]; then
+              echo "Command timed out. Retrying in $RETRY_DELAY seconds."
+            elif [ $status -eq 143 ]; then
+              echo "Command terminated by SIGTERM signal."  
+            else
+              echo "Command timed out successfully for $i iteration."
```
mengqiy, line 197: Earlier comment is not addressed. It will fall into this else clause when exit code is 0. And it will produce confusing log message.

Ganiredi:
> 

Reverted accordingly

mengqiy (approved):

.github/workflows/kitctl-test.yml
```diff
@@ -0,0 +1,218 @@
+name: This workflow automates the build, testing, and bootstrapping of a test environment on Kubernetes using KIT CLI
```
mengqiy, line 1: This is too long for a GH action name. See how its GH ci badge renders in the README.
Maybe just call it `kit-e2e`.
Let's fix this in a followup PR

PR #372, opened 2023-02-21
Update kitctl to v0.0.22
Created by Github actions during Release CLI action

prateekgogia:


mengqiy (approved):

mengqiy:
Looking at the release page. It seems we only build binary for darwin_amd64. Is there a reason we do build for other os and arch?

PR #376, opened 2023-02-23
Use stronger matching for finding existing instance

mengqiy:
Checking launch template name matches besides checking version.

Kitctl may use other substrate's instance without this change.

Issue #, if available:

Description of changes:

### DIFF
--- a/substrate/pkg/controller/substrate/cluster/instance.go
+++ b/substrate/pkg/controller/substrate/cluster/instance.go
@@ -42,12 +42,10 @@ func (i *Instance) Create(ctx context.Context, substrate *v1alpha1.Substrate) (r
 	for _, reservation := range instancesOutput.Reservations {
 		for _, instance := range reservation.Instances {
 			if aws.StringValue(instance.State.Name) == ec2.InstanceStateNameRunning || aws.StringValue(instance.State.Name) == ec2.InstanceStateNamePending {
-				for _, tag := range instance.Tags {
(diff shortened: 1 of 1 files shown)


substrate/pkg/controller/substrate/cluster/instance.go
```diff
@@ -133,3 +131,15 @@ func (i *Instance) delete(ctx context.Context, substrate *v1alpha1.Substrate, pr
 	logging.FromContext(ctx).Infof("Deleted instances %v", aws.StringValueSlice(instances))
 	return nil
 }
+
+func tagsMatch(ec2Tags []*ec2.Tag, substrate *v1alpha1.Substrate) bool {
+	for _, tag := range ec2Tags {
+		if aws.StringValue(tag.Key) == "kit.aws/subtrate" && aws.StringValue(tag.Value) != substrate.Name {
```
hakuna-matatah, line 137: isn't this already filtering instances based on the tag you're looking for -  [here](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/376/files#diff-02ec6b860e5c6ddc13f877ae4bec3be0234e51ac83c05dad0c98dbc765afaab9R38)  with help of this [function](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/17c7b32fa3b383484d216590146aa6b7c0cfdabb/substrate/pkg/utils/discovery/tags.go#L40-L49) ?  Why do we want to re-check again ?

PR #389, opened 2023-03-17
add ELB tags to subnet

oliviassss:
Description of changes:
Add `kubernetes.io/role/elb` and `kubernetes.io/role/internal-elb` ELB tags to subnet, in order for AWS LBC to discover the subnets for testing purpose.

### DIFF
--- a/substrate/pkg/controller/substrate/infrastructure/subnets.go
+++ b/substrate/pkg/controller/substrate/infrastructure/subnets.go
@@ -29,6 +29,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	TagKeySubnetInternalELB = "kubernetes.io/role/internal-elb"
+	TagKeySubnetPublicELB   = "kubernetes.io/role/elb"
+)
+
 type Subnets struct {
 	EC2 *ec2.EC2
 }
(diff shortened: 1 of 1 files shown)


substrate/pkg/controller/substrate/infrastructure/subnets.go
```diff
@@ -97,11 +102,20 @@ func (s *Subnets) ensureSubnet(ctx context.Context, substrate *v1alpha1.Substrat
 		return describeSubnetsOutput.Subnets[0], nil
 	}
 	// tag required by ELB controller to discover these subnets to configure ELB
+	// TODO: add the public and private ELB tags to public or private subnets separately
```
prateekgogia, line 105: `subnetSpec.Public` can give us if the subnet is public and we can tag subnets accordingly. I would suggest lets do that as part of this PR.

prateekgogia (changes requested):

substrate/pkg/controller/substrate/infrastructure/subnets.go
oliviassss: Yes, but right now deveks only creates public subnets, so `subnetSpec.Public` will only return `true`. We want to add the `kubernetes.io/role/internal-elb` to the same public subnets just for LBC to discover the subnets during test.

prateekgogia (approved):
LGTM!

PR #399, opened 2023-04-25
Adds pipeline for aws-nuke

ashishranjan738:
This commit adds pipeline for aws-nuke to periodically delete aws resources.


tests/pipelines/cleanup/aws/binding.yaml
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: triggers.tekton.dev/v1alpha1
+kind: TriggerBinding
+metadata:
+  name: aws-account-cleanup
+  namespace: tekton-pipelines
+spec: 
+  params: 
+    - name: aws-nuke-s3-config-path
+      value: <Replace this with s3 path of config>
```
hakuna-matatah, line 10: let's move this to internal ? 
Because Flux will automatically apply this file and when it can't file this value, the job will start failing.

PR #400, opened 2023-04-27
refactor perfdash addon to automate deployment

mengqiy:
Issue #, if available:

Description of changes:
Refactor the perfdash addon to automate the serviceaccount creation and binding with the correct permission in AWS IAM Role to read from s3 CL2 results.
It reduce the manual work to wire these things together.
I removed the `FluxRepoAddonPaths` flag, since there will be no consumers.


infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,241 @@
...
+  - "perfDashPrefix: 2k test"
+  - "perfDashJobType: performance"
+  - "perfDashArtifactsDir: "
+  `
+    },
+  })
+  cm.node.addDependency(ns)
+
+  const deploy = props.cluster.addManifest('perfdash-deploy', {
+    apiVersion: 'apps/v1',
```
hakuna-matatah, line 78: same with this ?

infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,241 @@
...
+            "app.kubernetes.io/part-of": "perfdash",
+        },
+    },
+    data: {
+        "PERFDASH_LOG_BUCKET": "my-bucket-name",
+        "AWS_DEFAULT_REGION": "us-west-2",
+        "AWS_REGION": "us-west-2",
+        "AWS_ROLE_ARN": sa.role.roleArn,
+        "jobs.yaml": `periodics:
+- name: "kit-eks-1k"
```
hakuna-matatah, line 62: I feel like config is something that  varies for different customers, I'm not understanding the rationale to cdkfy this over just applying as yaml file through flux from the repo (or) worst case just leave it as is internally like we do today ?

infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,241 @@
...
+  tags:
+  - "perfDashPrefix: 2k test"
+  - "perfDashJobType: performance"
+  - "perfDashArtifactsDir: "
+  `
+    },
+  })
+  cm.node.addDependency(ns)
+
+  const deploy = props.cluster.addManifest('perfdash-deploy', {
```
hakuna-matatah, line 77: So far theme of the code is, we install addons that have dependency with AWS resources with CDK and addons without aws resource dependencies through just flux. 
What's your intent in cdkfying this over the current way which is being applied through flux ?

infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,241 @@
...
+        name: "perfdash-config",
+        namespace: props.namespace,
+        labels: {
+            "app.kubernetes.io/name": "perfdash",
+            "app.kubernetes.io/component": "config",
+            "app.kubernetes.io/part-of": "perfdash",
+        },
+    },
+    data: {
+        "PERFDASH_LOG_BUCKET": "my-bucket-name",
```
hakuna-matatah, line 57: Having this name like this is not going to be helpful for any customer unless we take this as an input param as part of CDK args otherwise when this config is applied by CDK this won't have any real value IIUC. Thoughts ?

infrastructure/lib/addons/perfdash.ts
mengqiy: As I mentioned in the PR description, new approach create an IAM Role with proper permissions and bind it to the k8s serviceaccount while the old approach expect user to create the Role with proper permission and then manually edit the k8s service account to enable the binding.

infrastructure/lib/addons/perfdash.ts
mengqiy: Everyone user may need different config in the configmap. I guess it makes sense to remove the configmap out of CDK while AWS_ROLE_ARN can be merged into the Deployment.

infrastructure/lib/addons/perfdash.ts
mengqiy: Responded in https://github.com/awslabs/kubernetes-iteration-toolkit/pull/400#discussion_r1179549785.
Even if we let user specify log bucket in flag, it'd be hard to do the same thing for `jobs.yaml`. A user managed configmap out of CDK would make more sense.

infrastructure/lib/addons/perfdash.ts
hakuna-matatah: I wasn't referring to the IRSA part here, just the deployment spec part of it. I'm fine with the IRSA part of it.

infrastructure/lib/addons/perfdash.ts
hakuna-matatah: Yup, that comment of mine came later after this comment when i was reviewing. Let's remove this out of the cdk part.

infrastructure/lib/addons/perfdash.ts
hakuna-matatah: Like for example today we install tekton tests policy sa account through CDK here https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L141-L162 but won't rest of the spec through flux.

hakuna-matatah: Or other way is to install through CDK but install through helm or yaml reference instead of having the spec completely embedded into the cdk code like for example here https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/addons/karpenter.ts#L63-L88

mengqiy:
Updated the PR to remove CM from CDK

infrastructure/lib/addons/perfdash.ts
mengqiy: > Like for example today we install tekton tests policy sa account through CDK here https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/kit-infrastructure.ts#L141-L162 but rest of the spec through flux.

When we move CM out, we have resources: namespace, serviceaccount, deployment, service in CDK.
SA depends on NS. Deployment depends on NS and SA, since it needs `sa.role.roleArn`. Svc seems to be the only resource can be moved out of CDK. But I don't feel it make much sense to do it for just one resource.

> Or other way is to install through CDK but install through helm or yaml reference instead of having the spec completely embedded into the cdk code like for example here https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/infrastructure/lib/addons/karpenter.ts#L63-L88

PerfDash is not distributed through helm. If we want to do it, we need to create and maintain a helm chart for it. IMO we may not want to maintain a chart when we don't want to get into the distribution business for PerfDash.

@hakuna-matatah WDYT?

infrastructure/lib/kit-infrastructure.ts
```diff
@@ -46,6 +47,8 @@ export class KITInfrastructure extends Stack {
       clusterName: id,
       vpc: vpc,
       role: clusterRole,
+      mastersRole: clusterRole,
```
hakuna-matatah, line 50: I have this change locally. If you are going with this change, you might also need to add `sts assume role` permissions for this role for the account where it's created using `composite IAM principal ` stuff.

infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,65 @@
...
+    const perfdashKustomizationManifest = props.cluster.addManifest('PerfdashKustomizationSelf', {
+      apiVersion: 'kustomize.toolkit.fluxcd.io/v1beta1',
+      kind: 'Kustomization',
+      metadata: {
+        name: 'flux-addon-perfdash',
+        namespace: props.namespace,
+      },
+      spec: {
+        interval: '5m0s',
+        path: "./infrastructure/k8s-config/clusters/addons/perfdash",
```
hakuna-matatah, line 52: The path for this is different from other addons that we install. Wondering if we could keep it consistent ? By either moving all of them under this folder or just move this to other folder and /or rename it ?

infrastructure/lib/kit-infrastructure.ts
mengqiy: I can drop this part.

infrastructure/lib/addons/perfdash.ts
mengqiy: This path contains yaml while other addons are all TS code. Do we really want to move them to the same place given they have different file types. @hakuna-matatah Thoughts?

infrastructure/lib/kit-infrastructure.ts
```diff
@@ -21,7 +22,7 @@ export class KITInfrastructure extends Stack {
     const installEBSCSIDriverAddon = this.getContextOrDefault("AWSEBSCSIDriverAddon", "true")
     const installKarpenterAddon = this.getContextOrDefault('KarpenterAddon', "true")
     const installKitAddon = this.getContextOrDefault("KITAddon", "true")
-    const repoAddonPaths = this.node.tryGetContext('FluxRepoAddonPaths')
+    const installPerfDashAddon = this.getContextOrDefault("PerfDashAddon", "false")
```
hakuna-matatah, line 25: how do we install other addons if you intend to replace this to just perfdash ?

infrastructure/lib/addons/perfdash.ts
```diff
@@ -0,0 +1,87 @@
...
+    const perfdashKustomizationManifest = props.cluster.addManifest('PerfdashKustomizationSelf', {
+      apiVersion: 'kustomize.toolkit.fluxcd.io/v1beta1',
+      kind: 'Kustomization',
+      metadata: {
+        name: 'flux-addon-perfdash',
+        namespace: props.namespace,
+      },
+      spec: {
+        interval: '5m0s',
+        path: "./infrastructure/k8s-config/clusters/kit-infrastructure/perfdash",
```
hakuna-matatah, line 52: have you tested this ? was it installing the addon onto the cluster ? because I only see a relative path defined here.

infrastructure/lib/addons/fluxv2.ts
```diff
@@ -128,34 +127,6 @@ export class FluxV2 extends Construct {
   });
   kustomizationManifest.node.addDependency(chart);
 
-  if (props.repoAddonPaths) {
```
hakuna-matatah, line 131: This gives customer the flexibility to install addons from multiple repositories that we are not  opinionated by us. Any reason why we want to remove this flexibility ?

infrastructure/lib/kit-infrastructure.ts
mengqiy: `FluxRepoAddonPaths` was introduced for perfdash and removing it won't impact other addons

infrastructure/lib/addons/perfdash.ts
mengqiy: Yes, I have tested it.

infrastructure/lib/addons/fluxv2.ts
mengqiy: `repoAddonPaths` was introduced for perfdash. And this flag is no longer need to install perfdash.
`repoAddonPaths` is basically doing the same as `kubectl apply -f <repoAddonPaths>`. And IMO it's more intuitive to use `kubectl apply -f`. e.g. applying some yaml file is more trivial with `kubectl apply` than running `cdk deploy` which sometimes give strange errors :/

hakuna-matatah: IIRC this is for flux to apply automatically addons from this folder - https://github.com/awslabs/kubernetes-iteration-toolkit/tree/main/infrastructure/k8s-config/clusters/kit-infrastructure 
Not just for perfdash. 
Or am i missing something here ?

infrastructure/lib/kit-infrastructure.ts
hakuna-matatah: I think it was introduced for not just perfdash but also for addons in this folder - https://github.com/awslabs/kubernetes-iteration-toolkit/tree/main/infrastructure/k8s-config/clusters/kit-infrastructure

hakuna-matatah (approved):

PR #401, opened 2023-04-27
increase default cpu/mem for karpenter provsioned instances

hakuna-matatah:
Issue #, if available:

Description of changes:
Increase the defaults for karpenter provisioned instances to avoid OOMs and make sure pods have enough CPUs to run workload.

### DIFF
--- a/infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
+++ b/infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
@@ -26,6 +26,9 @@ spec:
       - amd64
   - key: kit.k8s.sh/app
     operator: Exists
+  - key: "karpenter.k8s.aws/instance-cpu"
+    operator: In
+    values: ["16", "32", "48", "64"]
   - key: kit.k8s.sh/control-plane-name
     operator: Exists
   limits: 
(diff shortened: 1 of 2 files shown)


infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
```diff
@@ -26,6 +26,9 @@ spec:
       - amd64
   - key: kit.k8s.sh/app
     operator: Exists
+  - key: "karpenter.k8s.aws/instance-cpu"
```
mengqiy, line 29: It's interesting to see you set cpu constraint to prevent OOM. It works in this case, since it provides more memory than default instance type.
Ideally we should use instance -memory if it's supported by karpenter.

mengqiy (approved):
One minor comment. It doesn't have to be blocking.

infrastructure/k8s-config/clusters/kit-infrastructure/kit-provisioner.yaml
hakuna-matatah: Yes,  karpenter does support memory types, we do need at least around 8 cpus range purely from cpu perspective for workload,  but when I looked at the ec2 instance types before doing this change, I noticed consistently across `c,m,r etc` instance types, if we need at least 16G of instance memory, we have to start with 16 cpu, so I went with this for now. My plan is to definitely reiterate this later to more nuanced and also we may choose to  ignore all this to do something at the workload level as well but was running into some issues with karpenter while doing so and was following up w karpenter on that but didn't want us to be blocked in the interim.

PR #403, opened 2023-05-04
Fix kitctl-test workflow name

Ganiredi:
Issue #, if available:

Description of changes:

Made changes for workflow run name  with kit-e2e-test : https://github.com/awslabs/kubernetes-iteration-toolkit/pull/365#discussion_r1184380428

This pull request modifies the `kitctl` deletion command behavior. Previously, the loop would break when the command execution was successful, indicated by an exit status of 0. With this update, the loop will continue to run the command, even if it was successful in the previous iteration as sporadically resources are getting leaked.

This change allows the workflow to keep running the `kitctl` command despite successful completions, as long as the elapsed time doesn't exceed the specified maximum duration.

### DIFF
--- a/.github/workflows/kitctl-test.yml
+++ b/.github/workflows/kitctl-test.yml
@@ -1,4 +1,4 @@
-name: This workflow automates the build, testing, and bootstrapping of a test environment on Kubernetes using KIT CLI
+name: kit-e2e-test
 on:
   push:
     branches: [main]
@@ -192,7 +192,6 @@ jobs:
             if [ $status -eq 0 ]; then
               echo "Command executed successfully for $i iteration."
               COMMAND_COMPLETED=true
-              break
             elif [ $status -eq 124 ]; then
               echo "Command timed out. Retrying in $RETRY_DELAY seconds."
             elif [ $status -eq 143 ]; then
--- a/substrate/pkg/controller/substrate/cluster/address.go
+++ b/substrate/pkg/controller/substrate/cluster/address.go
@@ -16,12 +16,10 @@ package cluster
 
 import (
 	"context"
-	"errors"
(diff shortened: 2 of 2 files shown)


mengqiy:
> Previously, the loop would break when the command execution was successful, indicated by an exit status of 0. With this update, the loop will continue to run the command, even if it was successful in the previous iteration as sporadically resources are getting leaked.

If kitctl leaks resources while still exits with 0, we should fix kitctl.

Ganiredi:
> > Previously, the loop would break when the command execution was successful, indicated by an exit status of 0. With this update, the loop will continue to run the command, even if it was successful in the previous iteration as sporadically resources are getting leaked.
> 
> If kitctl leaks resources while still exits with 0, we should fix kitctl.

Reason `Handle "InvalidAssociationID.NotFound" error in Address Delete function` :

This new change  improves the robustness of the Address struct's Delete function by handling the "InvalidAssociationID.NotFound" error that occurs when attempting to disassociate an Elastic IP address with an association ID that does not exist or has already been disassociated.

Previously, when the kitctl command started deleting resources and encountered this error, the process would stop and force quit, requiring the user to rerun the kitctl command again to delete the remaining resources. This change ensures that the kitctl command can continue deleting resources without leaking any resources.

Ganiredi:
To ensure the updated Delete function effectively handles the "InvalidAssociationID.NotFound" error and prevents the kitctl command from stopping and force quitting, the workflows have been run more than 10 times without encountering the issue.

Reference: [GitHub Actions Workflows - kitctl-test.yml](https://github.com/Ganiredi/kubernetes-iteration-toolkit/actions/workflows/kitctl-test.yml)

mengqiy (approved):

PR #410, opened 2023-06-27
move tekton tasks and pipelines to scalability ns to avoid PSA enforcement at tekton-pipelines namespace level

hakuna-matatah:
move tekton tasks and pipelines to scalability namespace to avoid PSA enforcement at tekton-pipelines namespace level as we moved to 1.26 from 1.21 . PSA is a thing since 1.25

example of the breaking PSA change causing pipelines and tasks to break 

https://github.com/tektoncd/pipeline/blob/cd8a151e2b52818e9d42ce4fd03053a657d6659e/config/100-namespace/100-namespace.yaml#L22

Issue #, if available:

Description of changes:


mengqiy:
Why are we deleting the `infrastructure/package-lock.json` file?
From my understanding, this file is important for creating reproducible build.

infrastructure/lib/kit-infrastructure.ts
```diff
@@ -45,6 +46,7 @@ export class KITInfrastructure extends Stack {
       clusterName: id,
       vpc: vpc,
       role: clusterRole,
+      kubectlLayer: new KubectlLayer(this, 'kubectl'),
```
mengqiy, line 49: Why do we need this?

hakuna-matatah:
> Why are we deleting the `infrastructure/package-lock.json` file? From my understanding, this file is important for creating reproducible build.

I don't think we are deleting the file. Given it's a huge file, it's not loaded by default in the diff.

infrastructure/lib/kit-infrastructure.ts
hakuna-matatah: we need this to make sure `kubectl` is selected as the same version of the cluster, otherwise it will be defaulted to `1.20` and that will not respect version skew and cause issues.

mengqiy (approved):
LGTM

PR #415, opened 2023-07-06
Update Pipeline version and provide access to cluster with iam role

hakuna-matatah:
Issue #, if available:

Description of changes:
- Update Tekton pipeline version to fix the bug https://github.com/tektoncd/pipeline/issues/6726 we are seeing on our dashboards
- Add clusterrole to have system:masters access for cluster and also output the access commands through cfn output.

### DIFF
--- a/infrastructure/lib/kit-infrastructure.ts
+++ b/infrastructure/lib/kit-infrastructure.ts
@@ -1,3 +1,4 @@
+import * as cdk from 'aws-cdk-lib';
 import { aws_ec2 as ec2, aws_eks as eks, aws_iam as iam, Stack, StackProps, STACK_RESOURCE_LIMIT_CONTEXT, Tags } from 'aws-cdk-lib'
 import { SecurityGroup } from 'aws-cdk-lib/aws-ec2'
 import { Construct } from 'constructs'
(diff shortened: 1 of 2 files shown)


infrastructure/lib/kit-infrastructure.ts
```diff
@@ -34,7 +35,10 @@ export class KITInfrastructure extends Stack {
 
     // The IAM role that will be used by EKS
     const clusterRole = new iam.Role(this, 'ClusterRole', {
-      assumedBy: new iam.ServicePrincipal('eks.amazonaws.com'),
+      assumedBy: new iam.CompositePrincipal(
+        new iam.ArnPrincipal(`arn:aws:iam::${cdk.Aws.ACCOUNT_ID}:root`),
```
mengqiy, line 39: Have you tested this part? Are you sure we should use `root` instead of `admin`?

mengqiy (approved):
One minor comment. LGTM otherwise.

infrastructure/lib/kit-infrastructure.ts
hakuna-matatah: Yes, I tested this part. It gives the Admin to assume this role to access this cluster. Given its admin role, should be allowed to assume.

PR #422, opened 2023-09-13
Fixes KitInfra k8s Layer and adds VPC delete step in teardown.

ashishranjan738:
This commit fixes KitInfra K8s Layer and adds VPC delete step in teardown.

Issue #, if available:

Description of changes:

### DIFF
--- a/infrastructure/lib/kit-infrastructure.ts
+++ b/infrastructure/lib/kit-infrastructure.ts
@@ -8,7 +8,7 @@ import { AWSFluentBit } from './addons/fluent-bit-for-aws'
 import { FluxV2 } from './addons/fluxv2'
 import { Karpenter } from './addons/karpenter'
 import { KIT } from './addons/kit'
-import { KubectlLayer } from 'aws-cdk-lib/lambda-layer-kubectl'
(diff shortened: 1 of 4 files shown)


infrastructure/lib/kit-infrastructure.ts
```diff
@@ -8,7 +8,7 @@ import { AWSFluentBit } from './addons/fluent-bit-for-aws'
 import { FluxV2 } from './addons/fluxv2'
 import { Karpenter } from './addons/karpenter'
 import { KIT } from './addons/kit'
-import { KubectlLayer } from 'aws-cdk-lib/lambda-layer-kubectl'
+import { KubectlV26Layer } from '@aws-cdk/lambda-layer-kubectl-v26'
```
hakuna-matatah, line 11: why 26 ? can't we on the default and make sure our stack is upto date ?

ashishranjan738: the default kubectl was still old and the stack creation was failing because of some resource mismatch. I think this is a better way because we can control both the kubectl version and kubernetes cluster version now.

hakuna-matatah (approved):
Lgtm

PR #439, opened 2023-11-21
Schedule kube-state-pod on same node as Prom pod for resources

hakuna-matatah:
Schedule kube-state-pod on same node as Prom pod for resources to avoid pod restarts or scheduling issues for kube-state-metrics pod as it demands[ high resources ](https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/pkg/prometheus/manifests/exporters/kube-state-metrics/deployment.yaml#L29-L37)


ashishranjan738 (approved):
LGTM

PR #442, opened 2023-11-29
apparently prometheus needs more than 10Gi given we enabled more scra…

hakuna-matatah:
Need more memory than the default set here - https://github.com/kubernetes/perf-tests/blob/2c9dc6095a3eb0d083b8abfed5646a9e9b2e293d/clusterloader2/pkg/prometheus/manifests/prometheus-prometheus.yaml#L10 as we are scraping more stuff. Otherwise it gets OOM killed and CL2 tests starts failing.

Screenshot 2023-11-28 at 4 16 29 PM


mengqiy (approved):
LGTM

PR #447, opened 2023-12-04
Install EKS Pod Identity Agent on Test clusters for CNI to use IRSAV2

hakuna-matatah:
Description of changes:
- Install EKS Pod Identity Agent on Test clusters for CNI to use IRSAV2
- Increase coredns replica count to 1000 to see if we can get better performance for DNS lookup latencies


tests/tasks/generators/clusterloader/load-slos.yaml
```diff
@@ -84,11 +84,11 @@ spec:
...
       CL2_RATE_LIMIT_POD_CREATION: false
       # enable SLOs related CL2 vars
-      CL2_NETWORK_LATENCY_THRESHOLD: 1s
+      CL2_NETWORK_LATENCY_THRESHOLD: 20s
       CL2_ENABLE_VIOLATIONS_FOR_NETWORK_PROGRAMMING_LATENCIES: true
-      CL2_NETWORK_PROGRAMMING_LATENCY_THRESHOLD: 30s
-      CL2_ENABLE_DNSTESTS: true
-      CL2_USE_ADVANCED_DNSTEST: true
+      CL2_NETWORK_PROGRAMMING_LATENCY_THRESHOLD: 300s
+      CL2_ENABLE_DNSTESTS: false
```
ashishranjan738, line 90: Why we are turning this off ? also can you provide some context on how increasing `CL2_NETWORK_LATENCY_THRESHOLD` and `CL2_NETWORK_PROGRAMMING_LATENCY_THRESHOLD` will help us ? does this mean our latency is too high ?

tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
```diff
@@ -78,4 +78,12 @@ spec:
       # install csi drivers.
       kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=$(params.aws-ebs-csi-driver-version)"
       # TODO: Calculate replicas based on the cluster size going forward.
-      kubectl scale --replicas 20 deploy coredns -n kube-system
\ No newline at end of file
+      kubectl scale --replicas 1000 deploy coredns -n kube-system
```
ashishranjan738, line 81: 20 -> 1000 ? hoping its not over provisioning.  I think we had a backlog item to calculate it dynamically based on the node size, shall we implement it

hakuna-matatah: That's why we still have a To-do, this requires a discussion w N/w team too on what's the right number, I'm increasing now because we are like 10X worse w these numbers.

ashishranjan738 (approved):
/LGTM

PR #462, opened 2024-05-30
Remove wait time as we don't depend on VAS

hakuna-matatah:
Issue #, if available:

Description of changes:
Remove wait time as we don't depend on VAS and we can add it back later when we depend on it and its still not faster


mengqiy (approved):
LGTM

PR #470, opened 2024-08-09
set revisionHistoryLimit to 0 for coredns

mengqiy:
Issue #, if available:

Description of changes:
We observed that a mixture of old and new replicas can be created when we scale the coredns deployment. This eventually get reconciled to all new pods. But the endpoint can observe the mixture of old and new pods.
Doing this should be able to help.

### DIFF
--- a/tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
+++ b/tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
@@ -81,8 +81,10 @@ spec:
       kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=$(params.aws-ebs-csi-driver-version)"
       # TODO: Calculate replicas based on the cluster size going forward.
       # Patching the coredns not to get scheduled on the monitoring node.
+      # Set revisionHistoryLimit to 0 to avoid keeping older replicasets.
       kubectl patch deployment coredns --patch '{
         "spec": {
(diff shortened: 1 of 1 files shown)


tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
```diff
@@ -81,8 +81,10 @@ spec:
       kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=$(params.aws-ebs-csi-driver-version)"
       # TODO: Calculate replicas based on the cluster size going forward.
       # Patching the coredns not to get scheduled on the monitoring node.
+      # Set revisionHistoryLimit to 0 to avoid keeping older replicasets.
       kubectl patch deployment coredns --patch '{
         "spec": {
+          "revisionHistoryLimit": 0,
```
hakuna-matatah, line 87: scaling the pods after bringing up the nodes ? Like somewhere here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tasks/setup/eks/awscli-mng.yaml#L117-L123  ?

hakuna-matatah (approved):

tests/tasks/setup/eks/awscli-cp-with-vpc.yaml
mengqiy: That's another thing that we can try.
Currently we makes all changes to addons before creating nodes.
Let's try this first.

PR #488, opened 2025-02-22
Add AL2023 launch template task

cartermckinnon:
Description of changes:

This adds a task, `awscli-eks-cfn-launch-template-al2023`, that creates a launch template compatible with AL2023-based EKS node AMIs.

The launch template may be used for a managed nodegroup, with or without an AMI specified, or an unmanaged nodegroup (autoscaling group).

The launch template includes the ability to pass additional options to the `kubelet`'s config file, via the `KubeletConfig` parameter.

### DIFF
--- a/tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
+++ b/tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
@@ -1,0 +1,97 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: awscli-eks-cfn-launch-template-al2023
+  namespace: scalability
+spec:
+  description: |
+    Create an EKS CFN stack to output a launch template for AL2023-based nodes.
+    This Task can be used to create an EKS CFN stack that outputs a launch template.
+    The launch template may be used for a managed nodegroup with or without a custom AMI.
+  params:
+    - name: cluster-name
+      description: EKS cluster you want to create CFN stack for.
+    - name: stack-name
+      description: Stack name you want to spin.
+    - name: region
+      default: "us-west-2"
+      description: The region where the cluster is in.
+    - name: kubernetes-version
+      default: "1.32"
+      description: The EKS version to install.
+    - name: ng-cfn-url
+      description: The url of the CFN YAML/JSON to create CFN stack for NG launch template
+    - name: endpoint
+      default: ""
+    - name: kubelet-config
+      default: "{}"
+      description: "Kubelet config JSON (will be merged with default config)"
+    - name: ami
+      default: ""
+      description: The AMI ID (or SSM parameter) to use for the launch template. If not provided, the launch template will not specify an AMI.
+  workspaces:
+    - name: config
+      mountPath: /config/
+  stepTemplate:
+    env:
+      - name: KUBECONFIG
+        value: /config/kubeconfig
+  steps:
+    - name: create-launch-template
+      image: alpine/k8s:1.23.7
+      script: |
+        set -o xtrace
+        set -o errexit
+        set -o pipefail
+
+        ENDPOINT_FLAG=""
+        if [ -n "$(params.endpoint)" ]; then
+          ENDPOINT_FLAG="--endpoint $(params.endpoint)"
+        fi
+
+        curl -s $(params.ng-cfn-url) -o ./amazon-ng-cfn
+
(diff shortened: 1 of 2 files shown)


tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -0,0 +1,96 @@
...
+  params:
+    - name: cluster-name
+      description: EKS cluster you want to create CFN stack for.
+    - name: stack-name
+      description: Stack name you want to spin.
+    - name: region
+      default: "us-west-2"
+      description: The region where the cluster is in.
+    - name: kubernetes-version
+      default: "1.28"
```
hakuna-matatah, line 21: let's use the latest k8s version as default please.

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -0,0 +1,96 @@
...
+  namespace: scalability
+spec:
+  description: |
+    Create an EKS CFN stack to output a launch template for AL2023-based nodes.
+    This Task can be used to create an EKS CFN stack that outputs a launch template.
+    The launch template may be used for a managed nodegroup with or without a custom AMI.
+  params:
+    - name: cluster-name
+      description: EKS cluster you want to create CFN stack for.
+    - name: stack-name
```
hakuna-matatah, line 15: I think we can avoid this param and leverage cluster-name ? Usually we try to pass as less params as possible in pipeline def. wdyt ?

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -0,0 +1,96 @@
...
+      description: The region where the cluster is in.
+    - name: kubernetes-version
+      default: "1.28"
+      description: The EKS version to install.
+    - name: ng-cfn-url
+      description: The url of the CFN YAML/JSON to create CFN stack for NG launch template
+    - name: endpoint
+      default: ""
+    - name: kubelet-config
+      default: "{}"
```
hakuna-matatah, line 28: please let's add description for this ? It might be obvious for us, may not be for other consumers.

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -0,0 +1,96 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: awscli-eks-cfn-launch-template-al2023
```
hakuna-matatah, line 5: Can we merge this  task with already existing [task](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tasks/setup/eks/awscli-cfn-lt.yaml) ?  



Majority of the stuff is a repeat like [this](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/488/files#diff-09a9b23e589ffd628197632a0b67671b48cc00b38f4ebad7b95e65e5a8b6e274R47-R68) and [this](https://github.com/awslabs/kubernetes-iteration-toolkit/pull/488/files#diff-09a9b23e589ffd628197632a0b67671b48cc00b38f4ebad7b95e65e5a8b6e274R85-R96 )  ?


wdyt ?

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
cartermckinnon: Sure, 1.28 is in the other existing Task FYI.

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
cartermckinnon: The intention was for this to be a drop-in replacement for the existing LT task, so if this is removed I think it should be done in a later PR

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
cartermckinnon: Sure, copied the description from the CFN template.

tests/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
hakuna-matatah: Yeah, those are old ones :D, should be updated. Trying to avoid it at least for new ones :D

PR #491, opened 2025-03-03
Allow script to continue deleting other resources

gavinnhh:
Allow script to continue deleting other resources even if delete cluster failed.
Also add to wait for cluster deletion.

Issue #, if available:

Cluster teardown in tekton pipeline exit early due to cluster not found and hence never reached to delete VPC and other resources. Cluster delete step should not exit early or block the entire teardown workflow.

Description of changes:
Add set +e to allow the script to continue execution even if a command fails.
Add wait command to wait for cluster deletion in case VPC stack deletion fail due to resources like subnet still attached to cluster.

### DIFF
--- a/tests/tasks/teardown/awscli-eks.yaml
+++ b/tests/tasks/teardown/awscli-eks.yaml
@@ -27,6 +27,7 @@ spec:
   - name: delete-cluster
     image: alpine/k8s:1.23.7
     script: |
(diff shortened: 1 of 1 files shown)


mengqiy (approved):
LGTM

hakuna-matatah (approved):

PR #496, opened 2025-03-12
Fix typos and refine Pod spec

xdu31:
Issue #, if available:

Description of changes:

Fix typos and refine Pod spec


tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -315,7 +315,7 @@ spec:
     - name: namespace
       value: $(params.kubernetes-version)
     runAfter:
-    - generate-eks-pod-identity
+    - generate
```
hakuna-matatah, line 318: If we just simply move this here, we would risk loosing signal from CL2 test.

I think for PIA, you need to have one more of this task to emit metric for different namespace outside of CL2 load test.

hakuna-matatah (approved):

PR #503, opened 2025-04-09
Add support for self-managed nodegroups

shvbsle:
Issue #, if available:

Description of changes:
Adding support for self managed node groups and migrating the `awscli-eks-cl2loadtest-with-addons-slos` to using self managed node groups


tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -88,6 +88,11 @@ spec:
     default: "20m"
   - name: timeout-pia-pod-startup
     default: "5m"
+  - name: unmanaged-nodegroup-cfn-url
+    default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/asg_node_group.yaml"
+  - name: launch-template-ami
```
shvbsle, line 93: move this to al2023 task yaml

sycned with @hakuna-matatah offline

hakuna-matatah:
It appears in both your test runs, prometheus pods are unable to come up. Could you check it is because of missing taints it is not able to schedule those pods on to the monitoring nodes  ?

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
-  name: awscli-eks-cluster-teardown
+  name: awscli-eks-cluster-teardown-shvbsle
```
hakuna-matatah, line 5: I don't we should be using your alias in task names :) 

Typo ?

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
```diff
@@ -89,3 +89,41 @@ spec:
...
       aws cloudformation wait stack-delete-complete --region $(params.region) --stack-name $(params.cluster-name)
       echo "Stack deleted successfully!"
+  - name: awscli-delete-asg
+    image: alpine/k8s:1.23.7
+    script: |
+      #!/bin/bash
+      set -e
+      aws sts get-caller-identity 
+      # Stack ids for self managed node groups will have pattern <cluster-name>-nodes-<num>
+      STACK_IDS=$(aws cloudformation describe-stacks \
```
hakuna-matatah, line 99: is this the only way to identify these SMNG nodes to delete them ?
Do we not have any APIs on EKS exposed to delete them ??

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -88,6 +88,12 @@ spec:
...
+    default: ""
```
hakuna-matatah, line 94: we should not have a default here.
Have defaults at task level please.

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
shvbsle: Im in favor of deleting the stacks that spawn the ASG since that would clean up any other resource that was spawned with the stack. 

There an API for deleting ASGs however:
https://docs.aws.amazon.com/cli/latest/reference/autoscaling/delete-auto-scaling-group.html

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
shvbsle: when defaulted to empty then launch-template-ami value is inferred automatically for your kubernetes version.
Passing a non-empty value here would override the ami

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
hakuna-matatah: Ok, in that case we don't have EKS APIs then to delete this.

shvbsle:
> It appears in both your test runs, prometheus pods are unable to come up. Could you check it is because of missing taints it is not able to schedule those pods on to the monitoring nodes ?

@hakuna-matatah The task that fails to bring up prometheus pods is not owned by me so Im not sure which missing taints I'm looking for. Can you explain what the role of prometheus is here?

tests/tekton-resources/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -62,12 +65,28 @@ spec:
...
+        if [[ -z "$KVERSION" ]]; then
+            echo "Failed to get cluster version, using provided kubernetes-version: $(params.kubernetes-version)"
+            KVERSION="$(params.kubernetes-version)"
+        fi
+        
+        # resolve ami
+        if [[ -n "$(params.ami)" ]]; then
+            AMI_VALUE="$(params.ami)"
+        else
+            AMI_VALUE="resolve:ssm:/aws/service/eks/optimized-ami/$KVERSION/amazon-linux-2023/x86_64/standard/recommended/image_id"
```
hakuna-matatah, line 87: given this task is rechanged to use only for SMNG, lets change the description of the task.

tests/tekton-resources/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -79,15 +98,20 @@ spec:
...
           > parameters.json
+        
+        cat parameters.json
 
         if [[ "$STACK_STATUS" == "" ]]; then
             aws cloudformation create-stack \
                 --stack-name $STACK_NAME \
                 --template-body file://$(pwd)/amazon-ng-cfn \
                 --parameters file://$(pwd)/parameters.json \
+                --capabilities CAPABILITY_IAM \
```
hakuna-matatah, line 114: Please add a description for this.

tests/tekton-resources/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -36,6 +36,12 @@ spec:
   - name: nodegroup-prefix
     description: Prefix that needs to be appended to asg names.
     default: ""
+  - name: unmanaged-nodegroup-cfn-url
+    default: ""
+    description: URL for "unmanaged nodegroup" (AutoScaling group) CloudFormation template. If not specified, a managed nodegroup will be created.
+  - name: launch-template-name
+    default: ""
```
hakuna-matatah, line 43: let's remove the default if its empty.

tests/tekton-resources/tasks/setup/eks/awscli-mng.yaml
```diff
@@ -36,6 +36,12 @@ spec:
   - name: nodegroup-prefix
     description: Prefix that needs to be appended to asg names.
     default: ""
+  - name: unmanaged-nodegroup-cfn-url
+    default: ""
```
hakuna-matatah, line 40: same here.

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
```diff
@@ -89,3 +89,41 @@ spec:
       # Wait for the stack to be deleted
       aws cloudformation wait stack-delete-complete --region $(params.region) --stack-name $(params.cluster-name)
       echo "Stack deleted successfully!"
+  - name: awscli-delete-asg
```
hakuna-matatah, line 92: Explained offline - this is basically a repeat of the new task iiuc `awscli-asg-delet-yaml`

hakuna-matatah: let's keep the task separately if you can to delete ASG, so its decoupled from this tear down task, and add it to pipeline def

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -194,6 +227,10 @@ spec:
       value: key=monitoring,value=true,effect=NO_SCHEDULE
     - name: nodegroup-prefix
       value: monitoring-
+    - name: unmanaged-nodegroup-cfn-url
+      value: ""
+    - name: launch-template-name
+      value: ""
```
hakuna-matatah, line 233: why set this to empty here ? could we just not set it ? and let task handle it automatically.

tests/tekton-resources/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
```diff
@@ -30,6 +30,11 @@ spec:
     - name: ami
       default: ""
       description: The AMI ID (or SSM parameter) to use for the launch template. If not provided, the launch template will not specify an AMI.
+    - name: node-role-name
+      description: The name of the IAM role to use for the node's instance profile, specified in the launch template.
+    - name: launch-template-name
+      default: ""
```
hakuna-matatah, line 36: this doesn't have any significance iiuc. If so, please remove this.

tests/tekton-resources/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
shvbsle: Changing description to:
```
Create an EKS CFN stack to output a launch template for AL2023-based nodes.
    This Task can be used to create an EKS CFN stack that outputs a launch template.
    The launch template may be used for a self-managed nodegroup with or without a custom AMI.
```

tests/tekton-resources/tasks/setup/eks/awscli-mng.yaml
shvbsle: done

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
shvbsle: In this comment we decided to remove the defaults at a task-definition level:
https://github.com/awslabs/kubernetes-iteration-toolkit/pull/503#discussion_r2042586155

If no default is provided then the task-definition treats the parameter as a mandatory field and you are required to pass some value to it. We can either handle defaults at task-def level or at pipeline-def level. The pro of defining it at the pipeline definition level is that it comes with visibility of what is passed to the tasks.

tests/tekton-resources/tasks/setup/eks/awscli-cfn-lt-al2023.yaml
shvbsle: done

tests/tekton-resources/tasks/teardown/awscli-eks.yaml
shvbsle: The current pipelines have a final teardown step [here](https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml#L7) where all the clean ups happen. Decoupling this would mean that we'd have to maintain 2 "final" steps just for cleaning up ASGs.

I'm of the opinion that `awscli-delete-asg` should be a part of `teardown` step since thats where various resources like VPC, cluster, IAM roles etc are deleted. I will instead remove the standalone `awscli-asg-delete.yaml`

hakuna-matatah (approved):

PR #506, opened 2025-04-30
Add/fix obelisk/journal test

gavinnhh:
Issue #, if available:
Obelisk periodic test has been failing.

Description of changes:
The test failure occurred because the slos pipeline (https://github.com/awslabs/kubernetes-iteration-toolkit/blob/main/tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml) was recently updated, but the Obelisk test pipeline (which was originally a copy of the slos pipeline with an additional journal migration task) wasn't updated to match. To prevent similar issues in the future, we should incorporate the obelisk/journal based cluster creation directly into the slos pipeline.

This approach would result in a single, unified pipeline for both Obelisk and non-Obelisk tests, eliminating the need to maintain separate pipelines and reducing the risk of inconsistencies.

For example, if we pass dataplane-configuration, eksadm-s3-path and cw-metrics-namespace it will create a obelisk/journal based cluster. This new workflow will not break the existing jobs/workflow. 

TODO:
Once this gets merged and synced to scalability account KIT cluster. 
1 Update the pipelinerun cronjob internally for both obelisk and non-obelisk. 
2 Set alarm in CW in scalability account (namespace will be changed (ex. from 1.32 to 1.32-raft) so need to update ALL in scalability account)

Edit:
I'm closing the PR as we move this change into internal code package.

### DIFF
--- a/tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
+++ b/tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
@@ -96,6 +96,15 @@ spec:
     type: string
   - name: ng-cfn-url-monitoring
     default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/eks_node_group_launch_template.json"
+  - name: dataplane-configuration
+    default: ""
+    type: string
+  - name: eksadm-s3-path
+    default: ""
+    type: string
+  - name: cluster-type
+    default: "raft"
+    type: string
   tasks:
   - name: slack-notification
     params:
@@ -153,6 +162,12 @@ spec:
       value: $(params.cluster-name)
     - name: kubernetes-version
       value: $(params.kubernetes-version)
+    - name: dataplane-configuration
+      value: $(params.dataplane-configuration)
+    - name: eksadm-s3-path
+      value: $(params.eksadm-s3-path)
+    - name: cluster-type
+      value: $(params.cluster-type)
     retries: 3
     runAfter:
     - create-cluster-node-role
@@ -358,7 +373,7 @@ spec:
     - name: value
       value: $(tasks.generate-eks-pod-identity.results.datapoint)
     - name: namespace
-      value: eks-pod-identity-$(params.kubernetes-version)
+      value: eks-pod-identity-$(params.kubernetes-version)-$(params.cluster-type)
     runAfter:
     - generate-eks-pod-identity
     taskRef:
@@ -371,7 +386,7 @@ spec:
     - name: value
       value: $(tasks.generate.results.datapoint)
     - name: namespace
-      value: $(params.kubernetes-version)
+      value: $(params.kubernetes-version)-$(params.cluster-type)
     runAfter:
     - generate
     taskRef:
--- a/tests/tekton-resources/tasks/setup/eks/awscli-cp-with-vpc.yaml
+++ b/tests/tekton-resources/tasks/setup/eks/awscli-cp-with-vpc.yaml
@@ -30,6 +30,12 @@ spec:
   - name: aws-pod-identity-agent-version
     default: v1.3.5-eksbuild.2
     description: The release version for aws pod identity agent.
+  - name: dataplane-configuration
+    description: ""
+  - name: eksadm-s3-path
+    description: ""
+  - name: cluster-type
+    description: ""
   workspaces:
   - name: config
     mountPath: /config/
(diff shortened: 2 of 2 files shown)


tests/tekton-resources/tasks/setup/eks/migrate-to-journal.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: tekton.dev/v1
+kind: Task
+metadata:
+  name: migrate-to-journal
```
mengqiy, line 4: Let's move this to internal

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -96,6 +96,17 @@ spec:
     type: string
   - name: ng-cfn-url-monitoring
     default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/eks_node_group_launch_template.json"
+  - name: is-obelisk
+    default: "false"
+    type: string
+    description: "Indicates the cluster will be migrated to journal/obelisk for testing."
+  - default: s3://journal-migration/EksAdm_linux_amd64
```
mengqiy, line 103: It'd be good to move this default to the internal Tekton task.

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -96,6 +96,17 @@ spec:
...
+    name: eksadm-s3-path
+    type: string
+    description: "The s3 path to the EKSADM binary for journal migration. This binary has been uploaded to S3 bucket in 197575167141"
```
mengqiy, line 106: It'd be good to move this to the internal Tekton task.

tests/tekton-resources/tasks/addons/cw-metric.yaml
```diff
@@ -21,9 +21,17 @@ spec:
       description: "The unit of the metrics"
     - name: value
       description: "The value for the metric"
+    - name: is-obelisk
```
mengqiy, line 24: I'm not in favor of this change.
This task should be generic. You should not change it here. You should pass `namespace=<foo>-obelisk` from outside of this task

tests/tekton-resources/tasks/setup/eks/awscli-cp-with-vpc.yaml
```diff
@@ -58,7 +62,16 @@ spec:
       echo "securitygroup=$sg"
 
       if [ "$CREATED_CLUSTER" == "" ]; then
-        aws eks create-cluster --name $(params.cluster-name) --region $(params.region) --kubernetes-version $(params.kubernetes-version) --role-arn $SERVICE_ROLE_ARN --resources-vpc-config subnetIds=$subnets,securityGroupIds=$sg $ENDPOINT_FLAG
+        if [ -n "$(params.dataplane-configuration)" ] && [ -n "$(params.eksadm-s3-path)" ] ; then
+          apk update
+          apk add --no-cache libc6-compat # Needed by eksadm binary
```
mengqiy, line 67: FMU if the eksadm is statically linked (CGO=0), we will not need to do this.
Can you please confirm this?
If this works, we can update the binary first, then this will become simpler.

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -96,6 +96,14 @@ spec:
     type: string
   - name: ng-cfn-url-monitoring
     default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/eks_node_group_launch_template.json"
+  - name: dataplane-configuration
+    default: ""
+    type: string
+  - name: eksadm-s3-path
+    default: ""
+    type: string
+  - name: cw-metrics-namespace
```
mengqiy, line 105: IMO it's OK to have a parameter "is-obelisk" at the pipeline level.
My earlier comment is for Tekton task "CW addon" which is intended to be generic.

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -358,7 +370,7 @@ spec:
     - name: value
       value: $(tasks.generate-eks-pod-identity.results.datapoint)
     - name: namespace
-      value: eks-pod-identity-$(params.kubernetes-version)
+      value: eks-pod-identity-$(params.cw-metrics-namespace)
```
mengqiy, line 373: it can be `$(params.kubernetes-version)-$(params.is-obelisk)`

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -371,7 +383,7 @@ spec:
     - name: value
       value: $(tasks.generate.results.datapoint)
     - name: namespace
-      value: $(params.kubernetes-version)
+      value: $(params.cw-metrics-namespace)
```
mengqiy, line 386: Same here.

tests/tekton-resources/tasks/setup/eks/awscli-cp-with-vpc.yaml
```diff
@@ -39,7 +43,7 @@ spec:
       value: /config/kubeconfig
   steps:
   - name: create-eks
-    image: alpine/k8s:1.23.7
+    image: alpine/k8s:1.30.0
```
mengqiy, line 46: When you update, why not using the latest?

gavinnhh: my bd this was actually for my test so it's typo

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
gavinnhh: I will use $(params.kubernetes-version)-$(params.cluster-type), where cluster-type can be a string of "obelisk" or "raft"

tests/tekton-resources/tasks/setup/eks/awscli-cp-with-vpc.yaml
gavinnhh: I have updated the binary to be static executable. Thx

PR #507, opened 2025-05-01
WIP: Split tasks into separate MNG and SMNG tasks

shvbsle:
Issue #, if available:

Description of changes:

### DIFF
--- a/tests/tekton-resources/tasks/setup/eks/awscli-mng.yaml
+++ b/tests/tekton-resources/tasks/setup/eks/awscli-mng.yaml
@@ -6,8 +6,8 @@ metadata:
   namespace: scalability
 spec:
   description: |
-    Create an EKS nodegroup, managed or unmanaged, for a given cluster.
-    This Task can be used to create an EKS managed or unmanaged nodegroup for a given VPC Subnets, security groups and service role in an AWS account.
+    Create an EKS managed nodegroup for a given cluster.
+    This Task can be used to create an EKS managed nodegroup for a given VPC Subnets, security groups and service role in an AWS account.
   params:
   - name: cluster-name
     description: The name of the EKS cluster you want to spin managed nodegroups for.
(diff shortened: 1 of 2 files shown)


tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -88,14 +88,15 @@ spec:
...
-  - name: unmanaged-nodegroup-cfn-url
+  - name: self-managed-nodegroup-cfn-url
     default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/asg_node_group.yaml"
   - name: launch-template-ami
     default: ""
     description: "Launch template ImageId value, which may be an AMI ID or resolve:ssm reference. By default resolve to the lates AL2023 ami for cluster version"
     type: string
-  - name: ng-cfn-url-monitoring
-    default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/eks_node_group_launch_template.json"
+  - default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/eks_node_group_launch_template_al2023.yaml"
```
hakuna-matatah, line 97: Now that we have split these into two separate tasks, leave defaults at task level.
That way, we can keep changes to pipeline def as minimal as possible.

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -88,14 +88,15 @@ spec:
     default: "20m"
   - name: timeout-pia-pod-startup
     default: "5m"
-  - name: unmanaged-nodegroup-cfn-url
+  - name: self-managed-nodegroup-cfn-url
     default: "https://raw.githubusercontent.com/awslabs/kubernetes-iteration-toolkit/main/tests/assets/asg_node_group.yaml"
```
hakuna-matatah, line 92: same here.

hakuna-matatah:
Thanks for following up on this task.

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
shvbsle: Made these changes in the latest commit. 
Do note that this becomes slightly inconsistent with how other params such as `ng-cfn-url` are defined in the pipeline, which do have default values. The only caution here would be that by moving these defaults to task-level definition, we are losing top-level visibility on what is being used under the hood by these task.

hakuna-matatah: Yeah, we need to move those as well to task level defaults, to keep pipeline minimal. Main motivation is - basically to not have to change all downstream pipelines built using these tasks as and when we change the defaults.

PR #511, opened 2025-05-23
/12 subnets in vpc and reserve those CIDR for prefix

mengqiy:
Issue #, if available:

Description of changes:
- tests/assets/amazon-eks-vpc-big.json contains the VPC with /12 CIDR
- VPC setup task will reserve /13 for prefix in those /12 CIDR. The reason why we don't reserve the full /12 CIDR is that IPAMD in VPC CNI picks the subnet with max number of `AvailableIpAddressCount`. If we reserve full /12 for prefix, AvailableIpAddressCount will become 0 and IPAMD will skip it.


hakuna-matatah (approved):
Lgtm

PR #512, opened 2025-05-29
Calculate metrics of credential fetching from Pods & upload to s3

xdu31:
Issue #, if available:

Description of changes:

1. Change Pod spec to invoke Pod Identity in normal container instead of from initContainer
2. Push credential fetching metrics
3. Push credential fetching time range, sample count, p50, p90 and p99 to s3

### DIFF
--- a/tests/assets/eks-pod-identity/pod-default.yaml
+++ b/tests/assets/eks-pod-identity/pod-default.yaml
@@ -6,18 +6,82 @@ metadata:
     group: {{.Group}}
 spec:
   containers:
-  - image: registry.k8s.io/pause:3.9
-    imagePullPolicy: IfNotPresent
-    name: pause
-  initContainers:
-  - name: app-init
+  - name: app-with-awsapi
     image: public.ecr.aws/aws-cli/aws-cli:latest
     imagePullPolicy: IfNotPresent
+    env:
+      - name: CLUSTER_NAME
+        value: "{{.ClusterName}}"
     command:
       - sh
       - -c
       - |
+        AUTH_TOKEN=$(cat $AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE)
+        MAX_ATTEMPTS=7
+        INITIAL_DELAY=0.2  # 200ms
+
+        NAMESPACE="EKSPodIdentityScalabilityTests"
+        DIMENSION_NAME="ClusterName"
+        DIMENSION_VALUE=$CLUSTER_NAME
+        METRIC_LATENCY_NAME="CredentialFetchLatency"
+        PERIOD=300
+
+        METRIC_MAX_RETRIES=5
+        METRIC_RETRY_DELAY=1
+
+        start_epoch=$(date +%s%3N)
+        # fetch credentials
+        for i in $(seq 0 $((MAX_ATTEMPTS - 1))); do
+          if curl -S -H "Authorization: $AUTH_TOKEN" http://169.254.170.23/v1/credentials; then
+            end_epoch=$(date +%s%3N)
+            printf "Endpoint is reachable at try %d\n" "$i"
+
+            latency_ms=$((end_epoch - start_epoch))
+            latency_sec=$(awk "BEGIN { print $latency_ms / 1000 }")
+
+            # send CredentialFetchLatency metric
+            for ((j=1; j<=METRIC_MAX_RETRIES; j++)); do
+              aws cloudwatch put-metric-data \
+                --namespace "$NAMESPACE" \
+                --metric-name "$METRIC_LATENCY_NAME" \
+                --dimensions "$DIMENSION_NAME=$DIMENSION_VALUE" \
+                --value "$latency_sec" \
+                --unit Seconds && {
+                  echo "Metric CredentialFetchLatency sent successfully."
+                  break
+              }
+
+              if [ "$j" -lt "$METRIC_MAX_RETRIES" ]; then
+                echo "Attempt $j failed. Retrying in $METRIC_RETRY_DELAY seconds..." >&2
+                sleep $METRIC_RETRY_DELAY
+                METRIC_RETRY_DELAY=$((METRIC_RETRY_DELAY * 2)) # exponential backoff
+              else
+                echo "Failed to send metric CredentialFetchLatency after $METRIC_MAX_RETRIES attempts." >&2
+                exit 1
+              fi
+            done
+
+            break
+          fi
+
+          if [ "$i" -eq $((MAX_ATTEMPTS - 1)) ]; then
+            echo "Max attempts reached. Exiting with failure."
+            exit 1
+          fi
+
+          SLEEP_TIME=$(echo "$INITIAL_DELAY * (2 ^ $i)" | bc -l)
+          printf "Failed. Sleeping %.3f seconds before retry...\n" "$SLEEP_TIME"
+          sleep "$SLEEP_TIME"
+        done
+
+        # s3 api call
         while ! aws s3 ls; do
             echo "Waiting for S3 bucket access..."
         done
         echo "S3 bucket is accessible, proceeding."
+
+        # pause
+        while true; do
+            echo "Sleeping for 1 hour..."
+            sleep 3600
+        done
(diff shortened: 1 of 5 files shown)


tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
```diff
@@ -175,6 +178,68 @@ spec:
...
+      if [ -n "$(params.endpoint)" ]; then
+        ENDPOINT_FLAG="--endpoint $(params.endpoint)"
+      fi
+
+      CLUSTER_NAME=$(params.cluster-name)
+      NAMESPACE="EKSPodIdentityScalabilityTests"
+      DIMENSION_NAME="ClusterName"
+      DIMENSION_VALUE=$CLUSTER_NAME
+      METRIC_LATENCY_NAME="CredentialFetchLatency"
+      PERIOD=300
```
hakuna-matatah, line 193: this should be tunable parameter

tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
```diff
@@ -175,6 +178,68 @@ spec:
...
+
+      START_TIME=$(aws eks $ENDPOINT_FLAG --region $REGION describe-cluster \
```
hakuna-matatah, line 195: could you add comments on why you consider this as start time ?

Also, please add comments overall to make it more readable for future users/consumers on your team, especially wherever you are making assumptions.

tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
```diff
@@ -175,6 +178,68 @@ spec:
...
+      total_samples=$(echo "$response" | jq '[.Datapoints[].SampleCount] | add // 0')
+
+      cat <<EOF > eks_pod_identity_test_summary.json
+      {
+        "start_time": "$START_TIME",
+        "end_time": "$END_TIME",
+        "total_samples": $total_samples,
+        "p50": $p50,
+        "p90": $p90,
+        "p99": $p99
```
hakuna-matatah, line 239: I see you are printing the latency numbers, but i don't see you validating the p99 under `x` seconds. Am i missing something here ?

xdu31: There are 2 options:
1. fail the task on threshold breaching
2. integrate with alarm on threshold breaching

tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
```diff
@@ -175,6 +194,81 @@ spec:
...
       ls -larth
       aws s3 cp . s3://$S3_RESULT_PATH/  --recursive
+
+      # if p99 is equal to or more than 2 seconds, exit with failure
+      int_p99=$(echo "$p99" | awk '{printf "%d", $1}')
+      if [ "$int_p99" -lt 2 ]; then
+        echo "p99 is less than 2"
+      else
+        echo "p99 is 2 or more"
+        exit 1
```
hakuna-matatah, line 273: We don't want to exit and fail the test, we need to capture the test result like this  https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3f5523db946e53dfcdabde926e03f8c06a434792/tests/tekton-resources/tasks/generators/clusterloader/load-slos.yaml#L171-L177


and use that to emit result like how we do it here - https://github.com/awslabs/kubernetes-iteration-toolkit/blob/3f5523db946e53dfcdabde926e03f8c06a434792/tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml#L360

So it can used configure alarm and cut tickets to your team.

tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
```diff
@@ -175,6 +194,81 @@ spec:
...
+        --namespace "$NAMESPACE" \
+        --metric-name "$METRIC_LATENCY_NAME" \
+        --dimensions Name="$DIMENSION_NAME",Value="$DIMENSION_VALUE" \
+        --start-time "$START_TIME" \
+        --end-time "$END_TIME" \
+        --period "$PERIOD" \
+        --statistics SampleCount \
+        --output json)
+
+      total_samples=$(echo "$response" | jq '[.Datapoints[].SampleCount] | add // 0')
```
hakuna-matatah, line 249: We need to capture what  is the `rate` at which credentials are fetched from service. To be able to compare that against the scheduler throughput.

Currently you have some kind of measuring by setting the timeout  https://github.com/awslabs/kubernetes-iteration-toolkit/pull/512/files#diff-d2d660edac904aa96e330bfae7bf67ef6885190877c5ab7668f0f157057da03fR61 accordingly by computing total number of pods and client pod creation rate/scheduler rate .

This will only give some kind of approximation  but not fully accurate throughput of your service using this kind of measurement.

tests/assets/eks-pod-identity/pod-default.yaml
```diff
@@ -6,18 +6,88 @@ metadata:
...
+            exit 1
+          fi
+
+          SLEEP_TIME=$(echo "$INITIAL_DELAY * (2 ^ $i)" | bc -l)
+          printf "Failed. Sleeping %.3f seconds before retry...\n" "$SLEEP_TIME"
+          sleep "$SLEEP_TIME"
+        done
+
+        # s3 api call
         while ! aws s3 ls; do
```
kmala, line 84: i think its better to verify the role being used to make sure we are not using the instance role.
`aws sts get-caller-identity | grep <role_name>`

tests/assets/eks-pod-identity/pod-default.yaml
```diff
@@ -6,18 +6,88 @@ metadata:
...
+        METRIC_MAX_RETRIES=5
+        METRIC_RETRY_DELAY=1
+
+        # make 7 attempts on credential fetching with exponential retries, and calculate the time taken
+        # push metrics on time taken on credential fetching
+        # to minimize failure from cloudwatch metrics, add retries on put-metric-data
+        start_epoch=$(date +%s%3N)
+        # fetch credentials
+        for i in $(seq 0 $((MAX_ATTEMPTS - 1))); do
+          if curl -S -H "Authorization: $AUTH_TOKEN" http://169.254.170.23/v1/credentials; then
```
kmala, line 41: do we need to verify that we got a successful response? may be status code 2xx ?

tests/tekton-resources/tasks/generators/clusterloader/load-pod-identity.yaml
xdu31: We have the metrics, and can integrate with an alarm, this should be out of this PR. I still think we should give a failure signal, just from cognitive perspective, a successful run is being alarmed, it can cause confusion

tests/assets/eks-pod-identity/pod-default.yaml
xdu31: I'm a bit concerned about sts quota we consume running in this account

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
```diff
@@ -84,6 +84,18 @@ spec:
     default: "200"
   - name: cl2-uniform-qps
     default: "100"
+  - name: cl2-metric-dimension-name
+    description: "default metric dimension name"
+    default: "ClusterName"
```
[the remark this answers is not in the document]
xdu31: but when we make a runpipeline, that's on pipeline level? So we don't need to make another code change here to change the name

tests/assets/eks-pod-identity/pod-default.yaml
kmala: i understand about the limits but i think we need a way to make sure that the command is successful because of token by pod identity and not using the instance role.

xdu31: I added a comment on how we can make sure this test is working as intended without checking on the assumed role identity

tests/tekton-resources/pipelines/eks/awscli-cl2-load-with-addons-slos.yaml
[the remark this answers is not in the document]
hakuna-matatah: Yeah, but run pipeline will take task level defaults when not supplied.

hakuna-matatah (approved):

PR #582, opened 2026-04-13
fix: flatten folders for multiple iterations for perfdash viz

dheeraj-coding:
Issue #, if available:
Flatten per iteration folders to enable perfdash graphing

Description of changes:

Tested on `perflab` and observed new data getting persisted in paths like `1776114444-1/` where the prefix is the unix timestamp and the suffix is the iteration number for that PCP load-test run. 

### DIFF
--- a/tests/tekton-resources/tasks/generators/clusterloader/sustained-scheduler-throughput.yaml
+++ b/tests/tekton-resources/tasks/generators/clusterloader/sustained-scheduler-throughput.yaml
@@ -146,8 +146,16 @@ spec:
     workingDir: $(workspaces.results.path)
     script: |
       S3_RESULT_PATH=$(cat $(results.s3_result.path))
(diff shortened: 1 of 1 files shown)


tests/tekton-resources/tasks/generators/clusterloader/sustained-scheduler-throughput.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
-  name: sustained-scheduler-throughput
+  name: sustained-scheduler-throughput-2
```
hakuna-matatah: typo ?

dheeraj-coding: yes good catch, was testing locally

hakuna-matatah (approved):

PR #586, opened 2026-04-29
feat: add tekton operator

dheeraj-coding:
Description of changes:
Add tekton operator, tekton pipelines and configure with kustomization for gitops. This would allow us to automatically reconcile new changes and maintain state in perf tests cluster for Tekton

### DIFF
--- a/infrastructure/flux/tekton/bucket.yaml
+++ b/infrastructure/flux/tekton/bucket.yaml
@@ -1,0 +1,13 @@
+apiVersion: source.toolkit.fluxcd.io/v1
+kind: Bucket
+metadata:
+  name: tekton-operator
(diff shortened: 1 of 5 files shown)


infrastructure/flux/tekton/bucket.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: source.toolkit.fluxcd.io/v1
```
hakuna-matatah, line 1: could you do folder structure something like this - "infrastructure/flux/addons/<addon-name>" - to make it clear that these are infrastructure addons applied through flux etc

hakuna-matatah (approved):

PR #597, opened 2026-07-12
Add pruner config to TektonConfig and increase retention count of PipelineRuns

chithreshazad:
Codifies the pruner settings that were previously only applied as an operator default (keep: 100). Increases retention to 1000 PipelineRuns (~6-8 weeks at current run rates) for better historical visibility.

### DIFF
--- a/infrastructure/addons/tekton/config.yaml
+++ b/infrastructure/addons/tekton/config.yaml
@@ -70,6 +70,12 @@ spec:
                   args:
                   - --external-logs=http://tekton-log-server.tekton-pipelines.svc.cluster.local:3000
                   - --stream-logs=false
+  pruner:
+    disabled: false
+    keep: 1000
+    resources:
+    - pipelinerun
+    schedule: "0 8 * * *"
   hub:
     options: {}
   result:


hakuna-matatah (approved):

infrastructure/addons/tekton/config.yaml
```diff
@@ -70,6 +70,12 @@ spec:
                   args:
                   - --external-logs=http://tekton-log-server.tekton-pipelines.svc.cluster.local:3000
                   - --stream-logs=false
+  pruner:
+    disabled: false
+    keep: 1000
+    resources:
+    - pipelinerun
```
hakuna-matatah, line 77: this ensures taskruns associatd to pipelineruns stays intact, right ?

chithreshazad: Yes - only pipelinerun is in the resources list, so the pruner only acts on PipelineRuns directly. TaskRuns owned by a PipelineRun are cascade-deleted by Kubernetes when the parent is removed, but stay intact as long as the PipelineRun is within the keep window (1000). Standalone TaskRuns (not owned by a PipelineRun) are never pruned.

PR #599, opened 2026-08-11
Wait 120s after cluster is ACTIVE before connecting (awscli-cp-with-pcp)

ronaldngounou:
## What

Add `sleep 120` **after** `aws eks wait cluster-active` in the `awscli-eks-cluster-create-with-pcp-stack` task, so subsequent steps don't connect to the cluster API server endpoint immediately after creation.

## Why

Loadtests intermittently failed during connect with:

```
couldn't get current server API group list: Get "https://<id>.sk1.beta.us-west-2.clusters.wesley.amazonaws.com/api?timeout=32s": dial tcp: lookup <id>...: no such host
```

This is a transient issue right after creation: the API server endpoint instances take time to come up and be discoverable, and the Route53 record has a ~60s TTL. The fix is to wait until the cluster create is signaled successful (`wait cluster-active`) and then give the endpoint time to become resolvable before kubectl connects.

### DIFF
--- a/tests/tekton-resources/tasks/setup/eks/awscli-cp-with-pcp.yaml
+++ b/tests/tekton-resources/tasks/setup/eks/awscli-cp-with-pcp.yaml
@@ -61,6 +61,11 @@ spec:
         aws eks create-cluster --name $(params.cluster-name) --region $(params.region) --kubernetes-version $(params.kubernetes-version) --role-arn $SERVICE_ROLE_ARN --resources-vpc-config subnetIds=$subnets,securityGroupIds=$sg --control-plane-scaling-config tier=$(params.control-plane-tier) $ENDPOINT_FLAG
       fi
       aws eks $ENDPOINT_FLAG --region $(params.region) wait cluster-active --name $(params.cluster-name)
+      # Cluster is ACTIVE (create succeeded). Wait for the API server endpoint to
+      # become resolvable/discoverable before connecting with kubectl: the endpoint
+      # instances take time to come up and the Route53 record has a ~60s TTL, so
+      # connecting immediately can fail with "no such host".
+      sleep 120
   - name: write-kubeconfig
     image: alpine/k8s:1.35.0
     script: |


tests/tekton-resources/tasks/setup/eks/awscli-cp-with-pcp.yaml
```diff
@@ -59,6 +59,8 @@ spec:
 
       if [ "$CREATED_CLUSTER" == "" ]; then
         aws eks create-cluster --name $(params.cluster-name) --region $(params.region) --kubernetes-version $(params.kubernetes-version) --role-arn $SERVICE_ROLE_ARN --resources-vpc-config subnetIds=$subnets,securityGroupIds=$sg --control-plane-scaling-config tier=$(params.control-plane-tier) $ENDPOINT_FLAG
+        # Give the control plane time to register before polling for ACTIVE
```
hakuna-matatah: isn't that the purpose of `wait` as well ?  the next command ?

ronaldngounou: Thanks. Updated by adding the `sleep 120` after receiving the signal that the cluster is active.

tests/tekton-resources/tasks/setup/eks/awscli-cp-with-pcp.yaml
```diff
@@ -61,6 +61,11 @@ spec:
         aws eks create-cluster --name $(params.cluster-name) --region $(params.region) --kubernetes-version $(params.kubernetes-version) --role-arn $SERVICE_ROLE_ARN --resources-vpc-config subnetIds=$subnets,securityGroupIds=$sg --control-plane-scaling-config tier=$(params.control-plane-tier) $ENDPOINT_FLAG
       fi
       aws eks $ENDPOINT_FLAG --region $(params.region) wait cluster-active --name $(params.cluster-name)
+      # Cluster is ACTIVE (create succeeded). Wait for the API server endpoint to
+      # become resolvable/discoverable before connecting with kubectl: the endpoint
+      # instances take time to come up and the Route53 record has a ~60s TTL, so
+      # connecting immediately can fail with "no such host".
+      sleep 120
```
hakuna-matatah: have you checked why `wait` won't  internally wait until endpoint reachable ?

ronaldngounou: `aws eks wait cluster-active` won't wait until the endpoint is reachable because the waiter model will only evaluate whether the cluster is active, failed, or deleting. It doesn't wait until the endpoint is reachable.

hakuna-matatah: please ensure the check is - loop until it becomes active and break and/or 120sec if that's the SLO for r53 recs.

ronaldngounou: Replaced the blind `sleep 120` with an active poll in the `write-kubeconfig` step: it loops on `kubectl get --raw=/readyz` (which exercises DNS + TCP + TLS + apiserver in one call, so it stays failing on the `no such host` DNS error), breaks as soon as the endpoint responds, and caps the wait at 120s (Route53 record TTL/SLO), failing the step if still unreachable at the deadline. It ends with a dummy `kubectl` call so the step is only marked successful once connectivity is verified.

tests/tekton-resources/tasks/setup/eks/awscli-cp-with-pcp.yaml
```diff
@@ -69,6 +69,26 @@ spec:
         ENDPOINT_FLAG="--endpoint $(params.endpoint)"
       fi
       aws eks $ENDPOINT_FLAG update-kubeconfig --name $(params.cluster-name) --region $(params.region)
+
+      # The cluster is ACTIVE, but the API server endpoint may not be resolvable yet:
```
chithreshazad, line 75: This checking of API Server endpoint should be done in its own step. Current step should only write the kubeconfig as per its name.

ronaldngounou: Moved it in its own step: `wait-for-api-server-endpoint`

chithreshazad (approved):

